PR #37, opened 2015-03-10
Checkpoints for every 2500 or so blocks from 50000

minercraig:
This gets rid of the Checkpoint warning.


BoXHocK:
I thought this was already fixed, but the release was not updated? It should work if you compile from source.

minercraig:
The visual checkpoint warning was removed, but you are still doing hardened checks.

IngCr3at1on:
Yeah hardended checkpoints are still in use (I'm not aware of a warning in relation to them) but updating them is never a bad idea anyway.

Looks good someone else will need to merge it though.

PR #39, opened 2015-03-20
Don't CheckProofOfStake on initial block download.

IngCr3at1on:
Returning false after initial block download causes a stal for most users.

Instead of removing the false return only run check if this is not the
initial block download.


IngCr3at1on:
Need to get someone else to test this one please.

IngCr3at1on:
Unfortunately I'm not sure this is quite working right...

Would like some other feedback before I close this request though.

IngCr3at1on:
Right... I figured out why this isn't quite working properly. Will need to look into the matter some more...

PR #47, opened 2015-04-03
Reset stake rate

IngCr3at1on:
As discussed on the dev calls this will lower or raise the stakerate on all existing primenodes to 100%
The keys are still grouped under their previous rates to insure proper handling of old blocks. To that end the conf files should not be changed from their existing rates. These will be forced to 100 by the code after the given date.

(we used timestamp over block number because timestamp allows us to set a time for mandatory wallet upgrades, all wallets agree on the time when verifying new blocks anyway).

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -12,10 +12,12 @@
 #include "util.h"
 #include "ui_interface.h"
 #include "checkpoints.h"
+#include "version.h"
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
 #include <boost/filesystem/convenience.hpp>
 #include <boost/interprocess/sync/file_lock.hpp>
+#include <ctime>
 
 #ifndef WIN32
 #include <signal.h>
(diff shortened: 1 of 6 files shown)


IngCr3at1on:
We have another PR we'll be putting in for the stall issue. It should probably have priority over this.

IngCr3at1on:
Since this will be a mandatory wallet update the timestamp on this will likely need to be expanded before merging to give people time to upgrade.

mjkoerner:
I'd like to see it even less,  more like 50-75 percent. The less coins the better.

HyperCommNet:
Please Talk here to help the community: http://talk.paycoin.com/discussion/539/pull-request-47-reset-stake-rate/p1

IngCr3at1on:
This update implements a protocol change to insure that only updated wallets are connected when the stakerate is updated.

evanlucas:
LGTM

evanlucas:
@HyperCommNet LGTY?

PR #49, opened 2015-04-03
Fix gitian builds

machee:
Update gitian configurations with new Qt binary name and foundation repo.


IngCr3at1on:
This one is important for distributed binaries.

PR #54, opened 2015-04-06
Update resource:

IngCr3at1on:
These should have never been merged.

This replaces the paybase resources with paycoin resources...


jwrb:
+1 We need to remove the old graphics and replace them with the correct ones as defined by the Paycoin brand guide.

PR #57, opened 2015-04-07
removed paybase logos

mgrav:
removed all pay base logos and replaced with pay coin ones


mitchellcash:
I know some of these overlap with @IngCr3at1on recent commits but I think @mgrav has used the Standard Colours based off the branding information so I think that would be preferable.

https://paycoin.com/downloads/PaycoinBrand.pdf

As another note, what do we think about changing the splash screen to an actual wallet like @mgrav has proposed. I personally don't think it looks as clean or as professional as the one currently in place. I would propose we accept all changes but the splash screen for now.

mgrav:
They do match the branding colours. if u look at the branding document
(bottom right) I actually made it.  Also please suit an alternative to the
splash if u like but I don't want to see any paybase logos on a paycoin
wallet. Paybase is almost gone and it's branding shouldn't be on the
paycoin wallet in the first place

mitchellcash:
My alternative is the one already in place. Referring to PR #54 from yesterday. Maybe you missed it but it shouldn't have the PayBase logo anymore:
https://github.com/PaycoinFoundation/paycoin/blob/master/src/qt/res/images/paycoin_splash.png

mgrav:
Sorry must have missed it. As long as it's paycoin then it's all good for
me

mitchellcash:
All good, it was less than 24 hours ago! :)

@HyperCommNet pending your thoughts and discussion at [talk.paycoin.com](http://talk.paycoin.com/discussion/574/pull-request-57-update-to-the-images-of-the-qt-wallet) are we able to merge all but the splash screen in this PR

IngCr3at1on:
I never even noticed the paybase branding was in the wallet until someone pointed it out the other day at which point we set to replacing it; beyond that I'll admit I'm semi-indifferent as to what icons are used.

evanlucas:
I'm -1 on `src/qt/res/images/paycoin_splash.png` but the rest look good to me

izerian:
Splash screen matters just about the least, considering the wallet fixes

IngCr3at1on:
I agree with @evanlucas, I like the splash we merged already.

PR #58, opened 2015-04-08
Update key.cpp

HFRleto:
consensus: guard against openssl's new strict DER checks
It was bothering me for a while.

### DIFF
--- a/src/key.cpp
+++ b/src/key.cpp
@@ -356,11 +356,23 @@ bool CKey::SetCompactSignature(uint256 hash, const std::vector<unsigned char>& v
 
 bool CKey::Verify(uint256 hash, const std::vector<unsigned char>& vchSig)
 {
-    // -1 = error, 0 = bad sig, 1 = good
-    if (ECDSA_verify(0, (unsigned char*)&hash, sizeof(hash), &vchSig[0], vchSig.size(), pkey) != 1)
+    if (vchSig.empty())
         return false;
 
-    return true;
+    // New versions of OpenSSL will reject non-canonical DER signatures. de/re-serialize first.
+    unsigned char *norm_der = NULL;
+    ECDSA_SIG *norm_sig = ECDSA_SIG_new();
+    const unsigned char* sigptr = &vchSig[0];
+    d2i_ECDSA_SIG(&norm_sig, &sigptr, vchSig.size());
+    int derlen = i2d_ECDSA_SIG(norm_sig, &norm_der);
+    ECDSA_SIG_free(norm_sig);
+    if (derlen <= 0)
+        return false;
+
+    // -1 = error, 0 = bad sig, 1 = good
+    bool ret = ECDSA_verify(0, (unsigned char*)&hash, sizeof(hash), norm_der, derlen, pkey) == 1;
+    OPENSSL_free(norm_der);
+    return ret;
 }
 
 bool CKey::VerifyCompact(uint256 hash, const std::vector<unsigned char>& vchSig)


IngCr3at1on:
That's funny, I have this in a branch I just keep forgetting about it cause of other issues.

HFRleto:
Yep, it's not critical, but it's been updated everywhere for some times now xD

IngCr3at1on:
Agreed, only thing that was holding me back is I really know nothing about DER checks in the first place and wasn't sure how to go about testing it.

HFRleto:
How can i make it happen ?

IngCr3at1on:
Do to the initial delay in the update we opted to not merge this prior to the update as we have not had time to fully test it.

That does not mean it will not be merged. We're attempting to make sure changes are tested in full before being merged. I am aware similar code has been pushed to most other coins but without comparing it I cannot speak to it's validity.

Please understand that this is not a matter of us not trusting you or your code it's simply a matter of wanting to make sure everything is checked multiple times before being merged.

HFRleto:
Ho no, im not mad about it not being merged !
i know you must have a schedule, but i can't access it :-(
My question was maybe not well written: my point was: i know i can't do the tests for you, since you have to check it yourselves, but it there any other way to help ?
Now i know that it's just a matter of time now, and that i can't help further on this matter.
i will move to the next fix.

IngCr3at1on:
I feel like we need a test method for this but I haven't been able to attempt to write one yet and there's not one in bitcoin do to the huge differences in the key.cpp code in the first place.

PaycoinMaster:
Closing this pull request.

HFRleto:
Why close this ?

IngCr3at1on:
Yeah I don't understand why this is closed either.... As I said it would be nice to test this but I wouldn't have closed it. @xmk3 @evanlucas thoughts? @PaycoinMaster what gives?

PaycoinMaster:
It was closed due to it being out of the ability to merge. And So far no tests have been Ran against the system. Once i test is Designed and this pull can be used we will reopen the ticket and test it and approve or disapprove the request.

IngCr3at1on:
Ahh... I wasn't aware we broke the ability to merge it lol.

Should we leave it open though in case @HFRleto wants to update the PR? With it closed he will have to open a new one after force pushing his branch github will not let him reopen this PR at that point while it would auto-update it if it were still open.

PaycoinMaster:
Very Well we will mark it.

HFRleto:
Sorry for the delay, but like IngCr3at1on, i wasn't aware i broke the ability to merge this !
I mean: "This pull request can be automatically merged by project collaborators." is pretty clear to me xD

IngCr3at1on:
Yeah @HFRleto I'm not really sure what @PaycoinMaster is referring to I noticed that after he re-opened it (perhaps confused it with another PR when writing that?)

IngCr3at1on:
Currently this change would not be effective/utilized; the following is an excerpt from the bitcoin 0.10.0 release notes (which can be viewed @ https://bitcoin.org/en/release/v0.10.0):

BIP 66: strict DER encoding for signatures

Bitcoin Core 0.10 implements BIP 66, which introduces block version 3, and a new consensus rule, which prohibits non-DER signatures. Such transactions have been non-standard since Bitcoin v0.8.0 (released in February 2013), but were technically still permitted inside blocks.

This change breaks the dependency on OpenSSL’s signature parsing, and is required if implementations would want to remove all of OpenSSL from the consensus code.

The same miner-voting mechanism as in BIP 34 is used: when 751 out of a sequence of 1001 blocks have version number 3 or higher, the new consensus rule becomes active for those blocks. When 951 out of a sequence of 1001 blocks have version number 3 or higher, it becomes mandatory for all blocks.

Backward compatibility with current mining software is NOT provided, thus miners should read the first paragraph of “Mining and relay policy enhancements” above.

----

Since the bitcoin code present in Paycoin is from version 0.6.3 this will not have any effect on the current wallet.

If the code is updated to require this functionality it would be prudent to fetch the following patch along with this one: https://github.com/bitcoin/bitcoin/commit/c6b7b29f232c651f898eeffb93f36c8f537c56d2

HFRleto:
You are right !
I saw Bitcoin 0.10 in the roadmap for may, but it was only for the  Wallet Port, not for a change in the transaction process !

PaycoinMaster:
Need to update alot for this to work.

PR #64, opened 2015-04-22
new beautiful GUI

BunBunny:
Default bitcoin GUI changed. New GUI looks better and uses paycoin brandbook


IngCr3at1on:
All the non-qt related code needs to be in a separate commit. This is so if something needs to be referenced later (to revert or otherwise modify) we can reference the commit in question without worrying about messing with other code that was committed alongside it.

For example I know the loop function in utils.h conflicts with qt5. IMHO this function should have never been written in the first place, it obfuscates the for(;;) function which is otherwise more than acceptable and recognized by most C++ programmers much more readily than "loop" (in short we're discussing changing that in a separate commit so this will need to be rebased assuming we do so). The same goes for general code cleanup, while this is fine it needs to be separated from the function that is being added/modified to make the commits with the actual additions/modifications easier to read.

jwrb:
The wallet does not follow brand guidelines correctly. Simply using colors from the guidelines does not constitute full adherence to them. Please use common sense when building interfaces. 

The guidelines can be found [here](https://paycoin.com/downloads/PaycoinBrand.pdf)

There are many places within the proposed UI where padding, buttons & text do not align correctly. Compared to the QT wallet this hurts usability across the board, and although it may not be a major issue it's another thing which is not perfect. 
We should also try and keep the wallet similar to the usual (classic) Qt style as used by most other third party wallets. This will keep the interface familiar & easy to use (especially true for users upgrading their wallet)

To keep the wallet similar, a horizontal navigation would be best. Instead of using a bright orange for selected tabs on grey, a combination of blue, white & a darkened (5-15%) blue would be best. Contrast must also be considered when inserting white text over an orange box; visually impaired or colour blind users may struggle with this.

If we are going to update the wallet, we should ensure that every single element within follows a similar visual style and that the wallet changes lower the entry point for new users of the coin. We can prettify things all we want, but until we do these things (from a branding & design perspective) we can never reach the level of adoption for the coin that we are all trying to work towards.

RdeWilde:
Cool.

Anyone got a screenshot of this?

BoXHocK:
+1 to @jwrb 

I personally feel like this is a too big change (visually) to be made into the official code. I'd want to see the official wallet using the GUI we all are familiar with, and this pull request as a fork only. That way people can decide themselves if they are ready for a change like this (on their own risk, that would be).

korfax:
Im all for making it look better than all other wallets. I respect what your trying to do.

mitchellcash:
+1 "All the non-qt related code needs to be in a separate commit."

BunBunny:
@RdeWilde  I've made an animated gif with screenshots for you
paycoin

@IngCr3at1on  Ok, I will split the code into several commits. I would need few days for this. 

@jwrb Paycoin guidelines do not cover design of an application. Existing guideline covers just very simple branding of website or paper document.
We can try to develop it and create design rules for applications. I've made the first iteration, we can improve on it together.
Regarding horizontal menu: firstly, I disagree that it is more convenient (regardless of what one feels it will be subjective either way); secondly, it is easier to add additional tabs with vertical menu – I was thinking of adding block explorer, trading bots, directly trading from the wallet (expanding the horizontal menu would probably render it to become cluttered). 
Regarding user experience – I am confident that end-users wouldn’t experience difficulties navigating the vertical menu as it is pretty simple.

BunBunny:
OMG, I’ve forgot to add an important style file in the commit: res/styles/default.qs. If you have managed to compile the wallet from the available code you would get an awful looking wallet. I am not even sure that it was possible to compile the wallet in the first place. I have uploaded the fix, please try to compile it now and let me know what you think.

IngCr3at1on:
Thanks for that @BunBunny I'll gladly have another look at the code once that is done. Please note we do have a commit in to replace the loop define with standard for(;;) statements though this has not been merged at this time.

RdeWilde:
@BunBunny Sweet! Good start! I am excited about the next iteration. Cutting it in small changes sounds great to me!

Suggestion; Show balance, open transaction, shortcut button (QR, wallet address) in the 'always visible' bar for quick access to most important stuff.

mitchellcash:
@BunBunny haha oh man that's why it looked so bad. Not looking into it too deeply I just pushed past the error and compiled and was thinking this is terrible haha. Looks much better now! I like it but I agree with comments above from @jwrb and @RdeWilde. With a few more iterations and refinements I would like to see this move ahead.

mitchellcash:
Maybe we can merge this into a new optional GUI branch so people can use what they prefer plus test and offer further refinements to this code until we are comfortable merging into master, or not.

BunBunny:
Fixed. Is it better?

IngCr3at1on:
Thanks for breaking up the code! :D

Any build changes needed?

PaycoinMaster:
Closing this pull request. But please update the README.md with a link to this Layout.

PR #65, opened 2015-04-24
refactor prime keys to separate file

IngCr3at1on:
This moves the prime node keys out of main.cpp (which is extremely crowded).
Should be considered a required step towards being able to add the prime mesh and micro primes.


IngCr3at1on:
@machee look familiar? :P

HyperCommNet:
Tested and Working :)

PR #66, opened 2015-04-24
remove loop define

IngCr3at1on:
loop causes a conflict with a Qt5 header


IngCr3at1on:
Aside from conflicting w/ the Qt5 headers the use of the "loop" define is obfuscating and generally not recommended. for(;;) is more appropriate for both readability as well as reducing the need for what some might call "evil defines".

xmk3:
This looks good as well.  @PaycoinMaster did we have a consensus on this one yet?

PR #70, opened 2015-04-27
Remove prime controller rates

whitj00:


sciborg:
I disagree with this. If you want to remove the prime controllers then you need to be able to pay the people that paid to have a PC like Adam Matlack known to pay over $500K for a PC.

whitj00:
@sciborg I think matlack might disagree

sciborg:
If thats the case then, I would be open to any suggestion that has a solid plan behind it and helps us the Paycoin community

korfax:
I think we need to keep the primes and lower the stake rate to no less than 25%

xmk3:
Unfortunately I believe your code changes would actually deadlock every wallet on the network instead of effectively removing the stake rate.  No issues here with removing the prime stake rates, but I think this would just break the coin outright.  :-)

@IngCr3at1on can probably verify that for me.

spydazcomputers:
ROFL kiss xpy goodbye if this gains any traction investors will be gone so fast HAHA we will be just another POS coin the stake rates are intended to pay for the Dev and backing needed to provide the features that are promised with this coin with out a profit margin to cover that expense

Finn-Sionnach:
There is no profit if the coins have no value, if they will still move transactions at 0% staking, this would be a good thing, at least until the coin goes up some. imho. Originally the prime controllers were meant to be flexible in the staking rate to help stabilize price, so let them do that. The people who have prime controllers would still benefit, over a little time. Adjust the staking rate every few months using supply/demand to help hold the price stable. Price fixing is very different from "No one is buying the blue slushies, let's make less of the blue syrup"

MitchellMint:
We need to remove the Primes and the Primes need to be compensated.  2.5 XPY for each 1 XPY the Primes own and then even PoS out.  Its a fair.  Just hard to organize and get donations for the 3 million XPY.  The destruction of coins is silly, and could have helped us in this situation.  Prime Owners will not renegotiate their position.

HFRleto:
I'm ok with it.
When the 6 month period is over, we should stop the primes for sometimes.
The day they actually do something useful we can add a higher stake rate.
But if we do it, we should remove them better, why keeping the key hardcoded if they stake at 5% ?
We should remove everything in relation with primenode rate, it will clear main.cpp a little xD

RdeWilde:
Why was this commit made, on what purpose? I just see "No description provided.".

Makes no sense just removing features/primes out of the blue by someone.

kmbriedis:
First, there is not point of having this in code 
```c++
} else if (nTime >= RESET_PRIMERATES) {
...
} else {
```

as the first IF statement `nTime >= RESET_PRIMERATES` will never ever be false again.

Also these lines are no-sense and I believe is coin breaker

```C++
             if (txNew.nTime >= RESET_PRIMERATES) {
                primeNodeRate = 5;
            } else if (txNew.nTime >= RESET_PRIMERATES) {
                 primeNodeRate = 100;
             }
```

HFRleto:
Why will it break the coin ?
But i agree, it's not a good way to remove the prime stake rate.

whitj00:
@xmk3 @kmbriedis I believe you're incorrect, can you explain the issue?

HyperCommNet:
The way the code was originally added. The only way to remove the primes over all is with a hard fork. Since we want to limit hard forks. A concept is in the work to make it so primes dont need the primary code line to function. If you would like to join i n on this concept we are creating a private channel on paycoin.slack.com for prime code talks

kmbriedis:
@whitj00 I am not completely sure of what I wrote, because I do not know were block is sent and where received

But just look at this line - it will never get to "else if", because both statements are same
```c++
if (txNew.nTime >= RESET_PRIMERATES) {
                primeNodeRate = 5;
            } else if (txNew.nTime >= RESET_PRIMERATES) {
                 primeNodeRate = 100;
             }
```

HFRleto:
@HyperCommNet, im interested in some prime code talks !

src/main.cpp
```diff
@@ -1420,7 +1420,10 @@ bool CTransaction::ConnectInputs(CTxDB& txdb, MapPrevTx inputs,
                     return DoS(100, error("ConnectInputs() : credit doesn't meet requirement for primenode = %lld while you only have %lld", MINIMUM_FOR_PRIMENODE, GetValueOut()));
                 /* Use time instead of block number because we can better
                  * control when a manditory wallet update is required. */
-                if (nTime >= RESET_PRIMERATES) {
+                if (nTime >= REMOVE_PRIMERATES) {
+                    if (nStakeReward > GetProofOfStakeReward(nCoinAge, 5) - GetMinFee() + MIN_TX_FEE)
+                        return DoS(100, error("ConnectInputs() : %s stake reward exceeded", GetHash().ToString().substr(0,10).c_str()));
+                } else if (nTime >= RESET_PRIMERATES) {
```
IngCr3at1on: Contrary to other peoples discussion we can't remove this second check because it's used when resyncing the blockchain so yes let's leave it but being as the first statement is already being evaluated to return an error.

Just a matter of personal preference (not saying it needs to be changed); I would use if instead of else if (it will still be checked in order of top-down) I've had some finicky behavior with else if on other projects. I expect this will work fine but I always feel like sharing random bits like that.

IngCr3at1on:
I have to disagree the w/ @xmk3, the code in place is pretty much the same as the reset-primerates code I don't believe it would break anything...

Having said that I would not want to do something like this until after we change the nMaxClockDrift in main.h (to my understanding this should be as low as possible without breaking staking functionality, as to the exact number I'm not sure yet). This will give less of a chance of nodes attempting to stake at 2 separate rates do to the clocks being off (something that was determined after the last update or I likely would have wanted it done first).

I'm debating if this would require a protocol update; one of the reasons we updated the protocol when resetting the prime rates was because in testing we had issues with non-updated nodes actually locking up the chain. This leads me to believe updating the protocol and (minimum protocol after the given update time) would need to be done for safety sake at the very least (this would essentially be another hard fork).

whitj00:
@IngCr3at1on obviously changes would be needed i.e. protocol.  This was to start a discussion

PR #72, opened 2015-05-01
bring package version inline with program

wdl1908:


IngCr3at1on:
We need to remove one of the debian package configs: currently there is one under contrib as well as one in the base repo (personally I'd prefer it under contrib as this matches the bitcoin repo structure), regardless of which one we retain we really do not need 2 debian package configs.

Being as I don't use the debian packages in anyway however I'm willing to entertain other opinions on this matter.

wdl1908:
Move the debian dir in the root to the contrib/debian then. but please keep the version up to date before you tag a release. This really is only so that debian packages for raspbian can be build.

IngCr3at1on:
Aye, understandable; I'll see about updating the release process ;)

IngCr3at1on:
@wdl1908 have a look at this before I put in a PR for it: https://github.com/IngCr3at1on/paycoin/tree/fix-up-debian-configs

PaycoinMaster:
Closing this pull request

PR #77, opened 2015-05-06
Add bash-completion script

IngCr3at1on:
slightly modified from bitcoin bash completion (tested on arch and ubuntu)
this is not installed w/ the debian package (yet)


BoXHocK:
I want to confirm this is fully working. Tested on Ubuntu.

PR #78, opened 2015-05-11
Fix translation filenames

mitchellcash:
* Match correct filename format to paycoin_xx_YY.ts
* Merge duplicate files


IngCr3at1on:
This includes translation updates as well (not listed in the commit comment); are these confirmed?

mitchellcash:
Thanks for the feedback @IngCr3at1on you're right.

Instead of writing 'remove duplicates' in the commit comment I probably should have written 'merge duplicates'. 

As some of the duplicate files had differences on certain lines I reviewed and went with what looked like the correct line. 

To be honest a big proportion of the translation changes were just removing the 'unfinished' translation tag from lines that were actually translated. Seems they were just left in by mistake.

I can review again today or if someone else would like to review and offer changes I'd be happy to have a look.

mitchellcash:
Sorry for the delay. Updated commit comment to reflect actual change of merging duplicating files rather than deleting duplicate files.

Trivial update.

PR #79, opened 2015-05-11
Update Doxygen logo

mitchellcash:
Not sure if we use Doxygen, but in case one day we decide to, the logo is now updated.

If we aren't going to use it we should just remove the image.


IngCr3at1on:
I know bitcoin has been using doxygen for quite awhile (though some comments have had to be updated to work w/ it which ours are probably not in line w/).

Personally I've never used it but that's not to say it wouldn't be a bad idea.

mitchellcash:
That's what I figured and even if we aren't using it now I'd still rather see a Paycoin logo in the docs folder rather than a Peercoin :)

mitchellcash:
Just pushed a small change where I ran the image through a Lossless Image Optimiser.

PR #88, opened 2015-05-21
Add new RPC command: `listaddressbook`

evanlucas:
Allows fetching the wallet’s address book. Passing `true` will return
the entire address book, not just wallet address.


IngCr3at1on:
I can confirm this works (watched Evan code and test it also confirmed on my linux box).

PR #91, opened 2015-06-20
Create new rpcnet module, and move 'getconnectioncount' RPC to it

mitchellcash:


IngCr3at1on:
I'm all for this, I've been wanting to restructure the code to more closely match bitcoin and this is a small step in that direction. We can move more over to this file after it's merged.

PR #92, opened 2015-06-20
Change nprev to pprev

mitchellcash:
It should be pprev because the next one is pnext and it's printing the memory address of the CBlockIndex (so pSomething).


IngCr3at1on:
Yep, that's fine.

PR #94, opened 2015-06-22
Prime node reroll

IngCr3at1on:


PaycoinMaster:
The mark has been set at 5% base and 25% prime.

IngCr3at1on:
@PaycoinMaster the numbers in the code are what Adam and @xmk3 and myself discussed on video call. Until I hear otherwise from more than 1 person I'm not changing them.

mitchellcash:
I note you've changed the rates back to 5% and 25%, but as a reference as to why these numbers were chosen, we were using the community vote as the basis of what the rates should be, from here:
https://talk.paycoin.com/discussion/1428/what-should-the-prime-rate-be/p1

That's where the 5% and 25% came from.

PR #109, opened 2015-06-26
XPY Splash screen and icon scale update

itsVictorVargas:
A few of these icons were bothering me after compering them to other wallets. I added higher resolution icon; fixed the splash screen since it was off center.

P.S. I am not sure if this is the proper way to go about doing this. Helping out with the UI, etc...


mitchellcash:
This is exactly the way to go about helping out with the UI etc. Keep it up!

Looks good to me.

IngCr3at1on:
If you guys say it looks better I'm for it. I don't see a difference in the splash screen or much of one in the paycoin icon itself but I also always use the daemon so meh.

itsVictorVargas:
The splash screen was only just 4 pixels off center, really not a big difference but it was just bothering me knowing that it was off center. If you use the swipe or onion skin tab you will see the difference :)

mitchellcash:
Yeah such a small difference but if the effort has been made to correct it, I'm happy with merging it in :)

PR #110, opened 2015-06-26
Fix some misspellings

mitchellcash:


IngCr3at1on:
I thought I was OCD but at least I've been able to convince myself to ignore that stuff. :P

mitchellcash:
I start working on something and notice one spelling mistake and then just have to go through everything :P
It also doesn't help I sometimes ```use git log --all --grep='spelling'``` or something similar to find spelling fixes in Bitcoin commits...

PR #118, opened 2015-07-04
Remove virtual keyboard from askpassphrasedialog and small clean-up

mitchellcash:
I want to get rid of the virtual keyboard from the askpassphrase dialog form. Most operating systems include their own on-screen keyboard and removing it helps simplify the interface.

Also helps clean-up the code a little bit.


itsVictorVargas:
I have to begin with a disclaimer: I am a romantic so my opinion might not be trustworthy. In OSX I think that the toggle keyboard looks quite sleek. If it looks sluggish on other operating systms and it's adding unessesary clutter to the code, then it might be time to remove it.

screen shot 2015-07-04 at 22 56 06

IngCr3at1on:
Shows how much I use qt... I didn't even know there was a toggle keyboard option.

Not sure I see a point in it considering anyone who doesn't have a keyboard is already going to have a virtual keyboard active.

src/qt/askpassphrasedialog.cpp
```diff
@@ -16,19 +16,15 @@ AskPassphraseDialog::AskPassphraseDialog(Mode mode, QWidget *parent) :
     fCapsLock(false)
 {
     ui->setupUi(this);
-
     ui->passEdit1->setMaxLength(MAX_PASSPHRASE_SIZE);
     ui->passEdit2->setMaxLength(MAX_PASSPHRASE_SIZE);
     ui->passEdit3->setMaxLength(MAX_PASSPHRASE_SIZE);
-
+
```
IngCr3at1on: Why the additional whitespace?

src/qt/askpassphrasedialog.cpp
```diff
@@ -240,10 +234,6 @@ bool AskPassphraseDialog::eventFilter(QObject *object, QEvent *event)
...
             }
         }
-    } else if (event->type() == QEvent::FocusIn) {
-        if (object->inherits("QLineEdit")) {
-            keyboard->setInput(object);
-        }
     }
     return false;
-}
+}
```
IngCr3at1on: While this is a bit of a legacy concern (most if not all compilers have been fixed to work anyway); used to be a new-line was required at the end of all files or else some compilers would crash.

As I said it's a legacy concern but it's also standard at this point to include a new line regardless.

mitchellcash: Good pickup, I must have gone and removed it in the original commit but I have now put it back.

I usually do try and remember to include the new line, as you said it may not be necessary anymore but it's still standard procedure. At least in this codebase.

src/qt/askpassphrasedialog.cpp
mitchellcash: Thanks, fixed.

mitchellcash:
Rebased.

Again, agree with @IngCr3at1on as anyone who doesn't have a physical keyboard is already going to have a virtual keyboard and most OS have this built in.

We don't need to include it as well.

itsVictorVargas:
Yup, makes sense. Cool beans :)

PR #119, opened 2015-07-04
Make user actions more straightforward in address book

mitchellcash:
- Add all actions that can be triggered with buttons to the context menu
- Hide delete action from context menu for receiving tab
- "Copy Address" instead of the vague, longwinded "Copy to Clipboard"

### DIFF
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -58,25 +58,34 @@ AddressBookPage::AddressBookPage(Mode mode, Tabs tab, QWidget *parent) :
         ui->signMessage->setVisible(true);
         break;
     }
-    ui->tableView->setTabKeyNavigation(false);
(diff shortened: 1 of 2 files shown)


IngCr3at1on:
Honestly I don't use the QT so other's opinions may matter more in this instance but I'd almost prefer "copy to clipboard"; it's not vague imho but standard (even github uses it for their clone commands on a repo). Be interested to know other peoples thoughts though.

mitchellcash:
Yeah I don't mind "Copy to Clipboard" either but it's how it is specifically used in the Qt wallet Address Book tab.

You click on an Address line in the tab which highlights multiple columns (Label and Address) and then press "Copy to Clipboard" but it's not clear by that whether you are copying Label, Address or both.

You are in fact only just copying the Address so it's best to update the button to better reflect the action, "Copy to Clipboard" is just too ambiguous as to what is actually about to happen.

PR #124, opened 2015-07-07
remove pre-paycoin protocol versions

IngCr3at1on:
Do NOT merge at this time : discuss

Peercoin had several proof of stake protocol versions that have never been active in Paycoin, this would remove those otherwise dead versions and their accompanied code.

We did this awhile back and I've been playing with it off and on. So far the only issue I've noticed is downgrading on a wallet that was synced using a binary with this code will cause a warning to appear in regards to it being past the upgrade time.

I consider this to be a code cleanup and would like others to test it and comment.

### DIFF
--- a/src/db.cpp
+++ b/src/db.cpp
@@ -520,16 +520,6 @@ bool CTxDB::WriteCheckpointPubKey(const string& strPubKey)
     return Write(string("strCheckpointPubKey"), strPubKey);
 }
 
-bool CTxDB::ReadV04UpgradeTime(unsigned int& nUpgradeTime)
-{
-    return Read(string("nProtocolV04UpgradeTime"), nUpgradeTime);
-}
-
-bool CTxDB::WriteV04UpgradeTime(const unsigned int& nUpgradeTime)
-{
-    return Write(string("nProtocolV04UpgradeTime"), nUpgradeTime);
-}
-
 CBlockIndex static * InsertBlockIndex(uint256 hash)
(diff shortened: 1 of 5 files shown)


mitchellcash:
Nice work @IngCr3at1on. I'm hoping to give this some testing over the next week.

IngCr3at1on:
Can't take full credit, I've been playing with it for awhile but if you look at the commit author I didn't do it lol

mitchellcash:
I've been doing a bit of testing with paycoin-qt and the only issue I get is downgrading to a wallet without this code clean up and I get the warning:
"WARNING: Blockchain redownload required approaching or past v0.4 upgrade deadline."

But I don't think that is an issue as preferably no one will be downgrading and if they do well they will just act on the warning and redownload the blockchain. No functionality to the wallet is getting lost so I think we should definitely move ahead with this.

If there is someone else out there who wants to run a few more tests I would be grateful. I didn't really look at paycoind yet.

IngCr3at1on:
Yeah that's the warning I was talking about.

IngCr3at1on:
Personally I'd like to merge this (I feel like the warning if you downgrade your wallet is minor); I'm curious if @evanlucas or @xmk3 have any thoughts or opinions first.

mitchellcash:
Completely agree :)

IngCr3at1on:
@PaycoinMaster any thoughts on this PR?

PR #136, opened 2015-07-25
qt: use series of pngs for spinner

mitchellcash:
Use a series of .png frames for the spinner instead of a .mng. `mng` is an obscure image format and is not built by default into Qt5 *(future planning)*.

This appears to improve the crispness of the spinner as well.

Does not noticeably increase the size (still ~27k) and the code is not more complicated either.


mitchellcash:
Also has a cool side effect of spinning at a speed relative to the speed of the blocks being downloaded.

Example:
5 blocks downloaded in 10 seconds - spins slower
10 blocks downloaded in 10 seconds - spins at twice the speed of 5 blocks in 10 seconds

IngCr3at1on:
Looks good to me, has my vote.

mitchellcash:
Thanks for testing @IngCr3at1on

PR #138, opened 2015-07-28
Fix minimum for primenodes to allow syncing the old primestakes

IngCr3at1on:
Currently syncing a new wallet fails when reaching the first prime stake from the original primenodes because it's minimum is below the defined value.

This will use the new value only if the nTime is after the end of phase one when running ConnectInputs.


IngCr3at1on:
Can we get this merged so people running off their own build can sync a new wallet directory?

PR #141, opened 2015-08-06
Fixes regression introduced in 17884411b7f754aaa56e9e8f0daba3838c0c06…

IngCr3at1on:
…85 that broke compiling on OS X


mitchellcash:
Good pickup! Sorry about that.

PR #145, opened 2015-08-08
Allow Qt to wrap long tooltips

mitchellcash:
Allow Qt to word-wrap tooltips. Qt can only word-wrap rich text messages.

Implemented without having to touch any translation: by listening for QEvent::ToolTipChange events, then rewriting the tooltips to prefix <qt/> and convert to rich text if it is not yet rich text.

No wrap: https://dl.dropboxusercontent.com/u/37902134/Paycoin%20Core/No%20wrap.png
With wrap: https://dl.dropboxusercontent.com/u/37902134/Paycoin%20Core/With%20wrap.png

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -1,12 +1,14 @@
 /*
  * W.J. van der Laan 2011-2012
- * The Paycoin Developers 2013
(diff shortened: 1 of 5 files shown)


IngCr3at1on:
Looks good to me (tested on my linux box).

PR #146, opened 2015-08-08
Cleanup of src/qt/forms/aboutdialog.ui

mitchellcash:
Initial cleanup of aboutdialog.ui. Further changes to this in mind I just want to get this base cleanup done and I'll have further pulls to follow.

### DIFF
--- a/src/qt/forms/aboutdialog.ui
+++ b/src/qt/forms/aboutdialog.ui
@@ -6,23 +6,27 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>544</width>
-    <height>488</height>
+    <width>593</width>
+    <height>319</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>About Paycoin</string>
   </property>
   <layout class="QHBoxLayout" name="horizontalLayout_2">
-   <property name="leftMargin">
-    <number>9</number>
-   </property>
-   <property name="topMargin">
-    <number>0</number>
-   </property>
-   <property name="bottomMargin">
-    <number>0</number>
-   </property>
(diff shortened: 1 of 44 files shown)


mitchellcash:
Rebased to include translation file fixes. Didn't actually look into the correctness of the translations just moved around to fit the change.

src/qt/forms/aboutdialog.ui
```diff
@@ -81,8 +59,11 @@
        </item>
        <item>
         <widget class="QLabel" name="versionLabel">
+         <property name="cursor">
+          <cursorShape>IBeamCursor</cursorShape>
+         </property>
          <property name="text">
-          <string notr="true">v0.1.0</string>
+          <string notr="true">0.3.666-beta</string>
```
IngCr3at1on: What is this version?

IngCr3at1on:
Builds and otherwise looks good (the about dialog actually fits in the window for a change).

I also have not confirmed validity of the translations.

src/qt/forms/aboutdialog.ui
mitchellcash: It's actually just used as preview text in Qt Designer.

I can change it back if needed but this seemed more appropriate. I just realised though I could have improved it yet again with something like 0.3.6.9-beta but to be honest it doesn't actually effect anything in production so I don't mind.

mitchellcash:
@IngCr3at1on my favourite part is it now fitting!

All translations need some work but that's not really the point of this pull :)

IngCr3at1on:
Yeah I'm fine with this, gets my vote.

PR #148, opened 2015-08-14
GUI: add 2 labels to the overviewpage that display Wallet and Transaction status

mitchellcash:
- cleanup overviewpage XML ui-file

Similar to the Bitcoin overviewpage so it's clearly visible from the UI that you can't expect transactions to show up before the blockchain download is complete.


IngCr3at1on:

PR #149, opened 2015-08-14
Remove string "TextLabel" from warningLabel as this is unneeded

mitchellcash:


IngCr3at1on:

PR #150, opened 2015-08-14
GUI: enhance translation possibilities in TransactionDesc / misc other changes

mitchellcash:
- remove "#include <QString>" as this is included in the header
- add some missing plural forms that can be translated
- change "yours" into "own address", which is easier to understand and
translate in that context
- cleanup translatable strings to not include HTML or unneeded chars
(e.g. ":")
- resize TransactionDescDialog a little (remove unwanted line-breaks
with non english translations)


IngCr3at1on:

PR #151, opened 2015-08-15
Should not be T minus, as this indicate duration to future event.

mitchellcash:
Change to a positive number, to indicate a past event.


IngCr3at1on:
Umm, not sure it really matters much imho but since it's just a print statement so long as it shows the right time I'm fine.

mitchellcash:
Yeah it matters very little. Sorry probably should have labelled trivial or something. I just liked the look and meaning without the minus better. Plus it saves space in the debug file ;) Very little space...

PR #152, opened 2015-08-15
Win32: use _strnicmp (ISO C++) instead of deprecated strnicmp (POSIX)

mitchellcash:
- the #ifdef WIN32 parts were moved to the top of the files as this is a place everyone looks from time to time
- add missing "#define strncasecmp _strnicmp" in init.cpp

See http://msdn.microsoft.com/en-us/library/ms235324%28v=vs.80%29.aspx for reference on _strnicmp.

Reference: https://github.com/bitcoin/bitcoin/commit/f4ac41806af5766199a7d526a7becbcb8a0f5ab3

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -21,6 +21,8 @@
 
 #ifndef WIN32
 #include <signal.h>
+#else
+#define strncasecmp _strnicmp
 #endif
 
 using namespace std;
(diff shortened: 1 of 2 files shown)


mitchellcash:
Don't merge this just yet. There is a small bug which I'll push a fix for shortly.

mitchellcash:
Fixed the bug, it should have been an ifdef WIN32, not an ifndef WIN32

**src/init.cpp**
from
```
#ifndef WIN32
#include <signal.h>
#define strncasecmp _strnicmp
#endif
```
to
```
#ifndef WIN32
#include <signal.h>
#else
#define strncasecmp _strnicmp
#endif
```

RdeWilde:
@mitchellcash that's the same code twice ;-)

mitchellcash:
Closed in favour of #154.

PR #157, opened 2015-08-17
make CheckDiskSpace() use a global static const uint64 for checking r…

IngCr3at1on:
…equired disk space and remove 2 ugly spaces from a message string

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2377,8 +2377,8 @@ bool CheckDiskSpace(uint64 nAdditionalBytes)
 {
     uint64 nFreeBytesAvailable = filesystem::space(GetDataDir()).available;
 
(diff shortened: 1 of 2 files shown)


IngCr3at1on:
I don't have space that is limited enough to cause this to occur but otherwise I'd consider this safe to merge.

Feel free to try to confirm if someone has a drive with enough stuff on it and/or a small enough drive.

mitchellcash:
We must have designed our message strings so nicely that you aren't actually removing an ugly spaces ;)

I can test this by making a small VM and filling up any empty space on the drive with junk data. But I would say this is safe to merge regardless.

IngCr3at1on:
Confirmed build in gitian.

PR #158, opened 2015-08-17
CTxMemPool: eliminate redundant lock, GetHash() call

IngCr3at1on:
::addUnchecked()'s only caller already takes the necessary lock,
and has already calculated the TX's hash.


IngCr3at1on:
This has been minorly tested on linux but I'd appreciate if someone else confirmed it before merging.

mitchellcash:
Is this something that can be merged into this pull as a second commit:
https://github.com/jgarzik/bitcoin/commit/96929a5515bb0ffee67aab7283b5ae9fcc1641d5

IngCr3at1on:
May as well (means my branch name is a bit of a misnomer but who cares).

PR #159, opened 2015-08-17
ensure that no double timestamps show up in the debug.log, by removin…

IngCr3at1on:
…g manual timestamps from the source (now only -logtimestamps parameter adds timestamps to debug.log)


IngCr3at1on:
I've had this running on 2 testnet nodes for an extended time, without reviewing every single line of the log that is supposed to be without timestamps I can't say with absolute certainty that this removes "all timestamps" with -logtimestamps disabled but I didn't see any so this appears to work exactly as intended.

PR #160, opened 2015-08-17
Alert system DoS prevention

IngCr3at1on:
This fixes two alert system vulnerabilities found by
Sergio Lerner; you could send peers unlimited numbers
of invalid alert message to try to either fill up their
debug.log with messages and/or keep their CPU busy
checking signatures.

Fixed by disconnecting/banning peers if they send 10 or more
bad (invalid/expired/cancelled) alerts.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3396,14 +3396,27 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv)
         CAlert alert;
         vRecv >> alert;
 
-        if (alert.ProcessAlert())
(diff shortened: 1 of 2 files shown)


IngCr3at1on:
We aren't directly utilizing alerts as is however it should still be possible for someone to send an invalid alert if they wanted to.

Safeguard against that as well as future proof a bit for if/when we start utilizing alerts.

This builds but I have not fully tested the functionality (it's current in bitcoin though so I'm not extremely worried on that front).

IngCr3at1on:
Confirmed build in gitian.

PR #161, opened 2015-08-18
GUI: start rows in askpassphrasedialog.ui at 0 (not 1)

mitchellcash:


IngCr3at1on:
Looks good, makes you wonder where the original developer learned how to count ;)

PR #165, opened 2015-08-19
Fix windows build

IngCr3at1on:
Revert the commits causing errors with gitian because I spent all of yesterday trying to fix it.

We will revisit this at a later time after compat has been cleaned up more.

PR #166, opened 2015-08-19
Prevent tooltip filter from ever causing infinite loops

IngCr3at1on:
I haven't seen this occur but I figured pull it in for good measure anyway (confirmed this builds on linux but I'd like to wait to merge it until I can test it in gitian which requires PR #165).


IngCr3at1on:
Confirmed build in gitian.

PR #171, opened 2015-08-23
Removed useless assignment

mitchellcash:
nLoadWalletRet is already equal to DB_NEED_REWRITE (we are in an if)


IngCr3at1on:
LMAO, so it is....

mitchellcash:
Haha I also laughed when I saw it. Very redundant.

PR #178, opened 2015-08-28
move rpc raw transaction functions out of bitcoinrpc to their own file.

IngCr3at1on:


mitchellcash:
An untested YAY! I can't wait until we split up all the RPC code.

mitchellcash:
Now a tested LGTM! Confirmed raw transactions via RPC can still be called.

PR #182, opened 2015-09-01
Reopen log file on SIGHUP

jwrb:
This PR implements the features found on [pr 917](https://github.com/bitcoin/bitcoin/pull/917) from Bitcoin. It causes a SIGHUP signal to make the daemon reload the log file, to facilitate log rotation.

I have tested this with logrotate, and all seems to be working (where the normal daemon would get stuck not writing to the log file)

This can be useful in situations where the daemon will write a large log file.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -77,6 +77,7 @@ bool fTestNet = false;
 bool fNoListen = false;
 bool fLogTimestamps = false;
 CMedianFilter<int64> vTimeOffsets(200,0);
+bool fReopenDebugLog = false;
 
 // Init openssl library multithreading support
 static boost::interprocess::interprocess_mutex** ppmutexOpenSSL;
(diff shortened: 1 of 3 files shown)


mitchellcash:
Thanks! I'm just on my phone so can't investigate much at the moment but what about including the 2nd commit as well: https://github.com/bitcoin/bitcoin/commit/fea25712ca54c4715f874cf3d871e84bbfd83352

src/util.cpp
```diff
@@ -223,6 +224,14 @@ inline int OutputDebugStringF(const char* pszFormat, ...)
             static boost::mutex mutexDebugLog;
```
jwrb: I would have, but it's already here @mitchellcash. We've already got that part in for some reason (Yay for serialised logs!)

mitchellcash: Ahh yes! I couldn't check very well on my phone. Good stuff.

IngCr3at1on:
Gitian builds ran, I can see you got the repeat boost thread out, thanks for that.

jwrb:
Great! As mentioned it compiles and runs as expected.

PR #184, opened 2015-09-01
Fix Up to date QT-Status

jwrb:
The status in the QT client was using a Bitcoin value of 90 minutes.

This should be more realistic as in the Paycoin world blocks are much faster, and happen every minute.

Dividing the amount of blocks behind by 10 (Bitcoin block time) gets us 9 minutes, which I consider an appropriate amount of time to be behind on blocks for Paycoin, although this is up for discussion.


IngCr3at1on:
Seems sane to me, all gitian builds run fine.

mitchellcash:
LGTM

PR #168, opened 2015-08-20
Windows: open debug.log file via Bitcoin-Qt

mitchellcash:
- This adds code to open (display) the debug.log on Windows with the associated application from the RPC console.
- Some small cleanups and tweaks.

Inspired by https://github.com/bitcoin/bitcoin/commit/4d3dda5d9f90d8aafb70c7e59beb27ec42d26790 but had to include a few other fixes.

PR #186, opened 2015-09-03
GUI fix: DebugLog file opens twice after clicking "Open" in RPC Console Information tab

mitchellcash:
Sorry I picked up on this in my original commit #168 but at the time it wasn't very detrimental and I didn't have the fix so I let it go through.

But here's the fix :)


IngCr3at1on:
Lol, someone will want to look at this in windows I guess :D

mitchellcash:
I've tested in Windows myself and it works. But happy for others to also test.

PR #188, opened 2015-09-03
Improve readability of CAddrInfo::IsTerrible

mitchellcash:
- Replaced 86400 with 24*60*60
- Remove references to specific timespans in comments

### DIFF
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -42,13 +42,13 @@ bool CAddrInfo::IsTerrible(int64 nNow) const
     if (nTime > nNow + 10*60) // came in a flying DeLorean
         return true;
 
-    if (nTime==0 || nNow-nTime > ADDRMAN_HORIZON_DAYS*86400) // not seen in over a month
+    if (nTime==0 || nNow-nTime > ADDRMAN_HORIZON_DAYS*24*60*60) // not seen in recent history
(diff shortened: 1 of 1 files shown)


IngCr3at1on:
This is debatable because obviously 86400 is 24*60*60

While the one may be slightly easier to read it's also using CPU cycles (nothing that would ever be noticeable of course lol); personally I'm fine with this either way.

mitchellcash:
Yeah it's nit picky for sure but a good reason I read was:



You must think in the context of changes over time, and how a change looks in a diff, not just the constant in isolation. Changing 24 to 20 hours is much more readable using
```
   - 24 * 60 * 60
   + 20 * 60 * 60
```
than
```
   - 86400
   + 72000
```
This notation is intentional. It creates self-documenting, maintainable code, that permits easier review of the time units being changed. Subtle, but IMO this is important to security critical code.

IngCr3at1on:
Personally I have no strong opinion here. I will mention that I changed something I did from `x * y * z` to `q` (where q is the product of the former) (basically the opposite of what you have done here) because it was in the prime node handling and was being ran a bunch of times for the P2 primes nodes and I felt reducing the CPU cycles was a good idea (I haven't pushed this yet for various reasons).

PR #189, opened 2015-09-03
Do not consider inbound peers for outbound network group exclusion.

mitchellcash:
Bitcoin will not make an outbound connection to a network group
(/16 for IPv4) that it is already connected to. This means that
if an attacker wants good odds of capturing all a nodes outbound
connections he must have hosts on a a large number of distinct
groups.

Previously both inbound and outbound connections were used to
feed this exclusion. The use of inbound connections, which can be
controlled by the attacker, actually has the potential of making
sibyl attacks _easier_: An attacker can start up hosts in groups
which house many honest nodes and make outbound connections to
the victim to exclude big swaths of honest nodes. Because the
attacker chooses to make the outbound connection he can always
beat out honest nodes for the consumption of inbound slots.

At _best_ the old behavior increases attacker costs by a single
group (e.g. one distinct group to use to fill up all your inbound
slots), but at worst it allows the attacker to select whole
networks you won't connect to.

This commit makes the nodes use only outbound links to exclude
network groups for outbound connections. Fancier things could
be done, like weaker exclusion for inbound groups... but
simplicity is good and I don't believe more complexity is
currently needed.

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -1375,16 +1375,17 @@ void ThreadOpenConnections2(void* parg)
         //
         CAddress addrConnect;
 
-        // Only connect to one address per a.b.?.? range.
+        // Only connect out to one peer per network group (/16 for IPv4).
         // Do this here so we don't have to critsect vNodes inside mapAddresses critsect.
         int nOutbound = 0;
         set<vector<unsigned char> > setConnected;
         {
             LOCK(cs_vNodes);
             BOOST_FOREACH(CNode* pnode, vNodes) {
-                setConnected.insert(pnode->addr.GetGroup());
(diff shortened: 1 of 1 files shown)


mitchellcash:
I've done some testing but if others could please also test.

IngCr3at1on:
I'd like to look at this one more but gitian builds run.

IngCr3at1on:
Aye this seems sane. Should probably deploy it on some nodes on the testnet and leave it running a few days (I'll see about doing that later today).

mitchellcash:
I left a single node running on testnet for a few days with now issue. Definitely welcome more nodes testing this though :)

PR #190, opened 2015-09-03
Update contrib/debian and remove system json_spirit patch.

mitchellcash:
That patch has been broken for a long, long time (forever?) and doesn't make sense.

Reference:
https://github.com/bitcoin/bitcoin/commit/3563824c602cb2b42cc21970935f45ce646c3964


IngCr3at1on:
Looks good, that patch hasn't ever been used with paycoin that much I know.

PR #191, opened 2015-09-03
Allow select()ing up to 1024 file descriptors on Windows

mitchellcash:


IngCr3at1on:
Seems sane. Any idea if there was a max before or just whatever the default is?

mitchellcash:
It was the just the default which was 64

PR #192, opened 2015-09-03
Prevent prevent file descriptor leak in ShrinkDebugFile()

mitchellcash:


IngCr3at1on:
Looks good; personally I'd have just put `if` instead of `else if` so that it just checks if the file is not null and closes it regardless of if the other function has run properly.

Course I'd have just written the first `if` slightly differently anyway and there would only be 2 calls to `fclose` instead of 3 lol

PR #193, opened 2015-09-03
Implement scrape addresses

IngCr3at1on:
Enable setting a scrape address for any in wallet address:
This will scrape all stake rewards at a transaction level and send the reward to
the scrape address.
Scrape addresses are stored in their own database and can be set, listed or
deleted using RPC via daemon or the in QT RPC console.

This could be added directly to the wallet.dat instead of using a separate database (meaning wallets may not be compatible with older client versions (untested)); for now until it's had more testing by other people, use a separate database.

RPC commands in action:
selection_106

Example transactions:
https://testnet.paycoin.com/transaction/3d8c25eeacd0261333e55044dbfd84d822a58deff7e5b5ab4652daf1fe2fc02b

https://testnet.paycoin.com/transaction/68beafcd22cccbf6aad005bee4068cc6ff30ec975edfc294824eef838dfc70bb

https://testnet.paycoin.com/transaction/b1dad54a1c43323f1ef2483c1e416ec7167eeaba6c282645b55f82dd79e6ef24

### DIFF
--- a/paycoin.pro
+++ b/paycoin.pro
@@ -129,6 +129,7 @@ HEADERS += src/qt/bitcoingui.h \
     src/script.h \
     src/init.h \
     src/mruset.h \
+    src/scrapesdb.h \
     src/json/json_spirit_writer_template.h \
     src/json/json_spirit_writer.h \
     src/json/json_spirit_value.h \
@@ -204,6 +205,7 @@ SOURCES += src/qt/bitcoin.cpp src/qt/bitcoingui.cpp \
     src/db.cpp \
     src/walletdb.cpp \
     src/primekeys.cpp \
+    src/scrapesdb.cpp \
     src/json/json_spirit_writer.cpp \
     src/json/json_spirit_value.cpp \
     src/json/json_spirit_reader.cpp \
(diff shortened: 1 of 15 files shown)


IngCr3at1on:
selection_108

mitchellcash:
So far, in my small testing, scraping to an external address seems to be working fine but there are some display issues that might need thinking about.

In the Transactions tab we aren't really highlighting where the Stake has gone. It can look a bit confusing.

For example, in the Address column on the Transactions page we put the address from which the stake came from but don't show any info as to where the Stake has gone. This now becomes a possible hole in the transaction record.

It is possible however when you view the transaction details of a Stake to see the address where you sent the stake, however, this now becomes confusing to the user because the Address Page is showing one address where the Transaction Details are showing another address.

PaycoinMaster:
Conflicts on this branch

mitchellcash:
Is there anymore discussion regarding my above comment?

IngCr3at1on:
@PaycoinMaster not surprised at all I've got all sorts of commits staged that have different conflicts lol...

@mitchellcash I agree with you however I don't think it's confusing enough to leave out but rather something we should look into fixing up as we go (just my personal opinion); I feel having the functionality is more important than whether it confuses someone.

IngCr3at1on:
@PaycoinMaster fixed conflict, good to go.

IngCr3at1on:
@mitchellcash let's open an issue for an improvement on this function regarding your comments (I can do this if you like).

mitchellcash:
Agree that we should merge and deal with nuances after in this case. Please see #210 for issue discussion.

PR #194, opened 2015-09-04
Refactor primenode key handling

IngCr3at1on:
This is step one in a major refactor of the primenode key handling.
It moves all the confirmation checks out of main.cpp and into primekeys.cpp as an extension of CTransaction.
Also removes primenode key signature checks on blocks generated on using the legacy signature OP codes; if a block passes all the chain integrity checks then we know it belongs so it's known good at this point.
There's some general fixup in IsPrimeStake on the actual check code.

This has been tested staking on the testnet and for syncing the main net but could still use some more confirmations.


mitchellcash:
You left `src/primekeys.h` in paycoin.pro

IngCr3at1on:
So I did, nice catch... (if that were a build breaking issue I'd have noticed lol)

Fixed.

PR #197, opened 2015-09-07
Add a testnet checkpoint at block 546

mitchellcash:
Introduce checkpoints to the testnet with a checkpoint at block 546.


IngCr3at1on:
Looks good, confirmed gitian builds and that this syncs properly on linux.

PR #198, opened 2015-09-08
Minor update to makefiles.

IngCr3at1on:
This has absolutely no effect on a standard build since the build.h file in
src/obj always lists the date and time of the last tag qt generates a build.h
file in build/ with the correct (modified) version which is used when creating
the version displayed in the binary but it is only generated by the QT (meaning
the daemon is modifying the date/time a different way).

But as it stands we're removing a not existent file on `make clean`, so let's
remove the correct file.


mitchellcash:
Looks good.

PR #200, opened 2015-09-08
Remove deprecated primenoderate option from CreateCoinStake.

IngCr3at1on:
This has been dead since the end of Phase One (not needed to sync).
Since all primenodes are the same rate they all use the same OP Code.


IngCr3at1on:
Has no-one looked at this (@PaycoinMaster, @mitchellcash)?

mitchellcash:
Sorry yep I have. It builds fine and syncs okay but haven't gone much past that testing wise until tonight :)

PR #203, opened 2015-09-11
Replace the alert key with something that we have a private key for...

IngCr3at1on:
I've been reading up on how this is handled in the code and from what I can tell we should be able to safely update this without any issues.

Unfortunately most clients won't have this key until the next release.

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -1806,7 +1806,7 @@ public:
     bool CheckSignature()
     {
         CKey key;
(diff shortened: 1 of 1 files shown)


mitchellcash:
Who will hold the alert key?

IngCr3at1on:
Currently I do (I also sent an encrypted file to @xmk3 with the key in it (just in case an F16 turbine falls out of the air and lands on top of me or something).

PR #204, opened 2015-09-12
remove "checkorder" and "reply" P2P commands

mitchellcash:
- These command are a leftover from send-to-IP transactions, which have been removed a long time ago.
- Also removes CNode::mapRequests and CNode::PushRequests, as these were only used for the mentioned commands.

Reference:
https://github.com/bitcoin/bitcoin/commit/529a4d4824136a00e63cd69831023b9c4dd6936b


IngCr3at1on:
I actually noticed checkorder wasn't being used Friday when I was looking at something else lol.

Looks good to me.

IngCr3at1on:
Merge conflict.

mitchellcash:
Rebased.

PR #205, opened 2015-09-13
Bring README more inline with current events

mitchellcash:
The README was very outdated with some broken links and inaccurate information. This not only brings it more up to date but also gives it a bit of a cleanup as well.


IngCr3at1on:
Looks good to me.

PR #207, opened 2015-09-13
P2P: Do not request blocks from peers with fewer blocks than us

mitchellcash:
If the remote node has a shorter chain, do not waste our
special getblocks request on them.


IngCr3at1on:
Seems sane, I wonder where Jeff got the number 144 lol

mitchellcash:
My best guess is each block time average is 10 minutes * 144 = 1440 minutes. Divide this by 60 minutes in an hour equals 1 day.

They didn't want to accept full block height as it could fail to fetch the new chain in some edge cases. So I guess they randomly chose 1 day behind being still okay to accept blocks from just because 1 day is a nice easy number haha.

This is my best guess as they don't actually mention how they calculated it but I'd say that's probably correct.

PR #208, opened 2015-09-14
CAddrDB: Replace BDB-managed addr.dat with internally managed peers.dat

IngCr3at1on:


mitchellcash:
I've been wanting this one to happen for a while, nice once! Tested and working perfectly :)

PR #209, opened 2015-09-14
Refactor primenode key handling 2

IngCr3at1on:
Second (final step) in major refactor of primenode key handling.

Creates a database for storing primenode control information (will also be used for microprimes); key, primenode rate, valid starting and valid until times. These can be used to store multiple sets of keys with different valid times (for easier handling of the primenode reroll each year).


jwrb:
Looks to be okay :+1:

IngCr3at1on:
@PaycoinMaster there you go.

PR #211, opened 2015-09-16
Make truncating the debug.log file an option.

jwrb:
Useful for developers who need to refer to futher back in debug.log history, but who don't want to
enable the -debug option and all the verbosity that comes with that.

This implements [1396](https://github.com/bitcoin/bitcoin/pull/1396)


IngCr3at1on:
Gitian built successful and works exactly as expected. LGTM

PR #213, opened 2015-09-17
Change Coin Control first column label

mitchellcash:
Not sure if this is merge worthy due to the fact I haven't seen this in the wild but I figure let's stop it before it starts and if you look at the latest Bitcoin master this patch still exists so I figure we should probably get it in.

### DIFF
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -104,6 +104,10 @@ CoinControlDialog::CoinControlDialog(QWidget *parent) :
     // (un)select all
     connect(ui->pushButtonSelectAll, SIGNAL(clicked()), this, SLOT(buttonSelectAllClicked()));
 
+    // change coin control first column label due Qt4 bug. 
+    // see https://github.com/bitcoin/bitcoin/issues/5716
+    ui->treeWidget->headerItem()->setText(COLUMN_CHECKBOX, QString());
+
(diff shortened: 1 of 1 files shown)


IngCr3at1on:
Most likely specific to Qt 4.8.6 (I haven't seen it with 4.8.7).
Seems sane either way and doesn't cause any issues (gitian built properly).

mitchellcash:
Sounds good, figure better safe than sorry if people are building with 4.8.6 :)

mitchellcash:
Rebased over master (plus I wanted Travis to run his eyes over it ;) )

IngCr3at1on:
I'm still partly on the fence for this PR, we haven't had this issue occur, ever.

mitchellcash:
We don't really have many people making their own builds so it is less likely. The bug does exist but it's up to you.

IngCr3at1on:
@mitchellcash that's just it though, does the bug really exist? Have you ever seen it in the wild?

I'm reminded of a tree falling in the forest, I can tell you it makes a sound but the simple fact of the matter is unless someone or something is there to hear it the truth of the matter is it makes no sound at all (sound is only sound if it's being picked up by something capable of translating the air waves into sound in the first place: IE, an ear drum). Very similar statements could be made about a bug that hasn't been confirmed :P

mitchellcash:
@ingCr3at1on haha I do like the expansion! Thanks for merging anyway.

PR #219, opened 2015-09-21
Show when node is misbehaving, not just at the point that it gets banned

jwrb:
Taken from [1343](https://github.com/bitcoin/bitcoin/pull/1343)

Compiles and runs fine. May require further testing with other nodes (that are misbehaving!)

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -586,7 +586,7 @@ bool CNode::Misbehaving(int howmuch)
 
     if (addr.IsLocal())
     {
-        printf("Warning: local node %s misbehaving\n", addrName.c_str());
+        printf("Warning: local node %s misbehaving (delta: %d)\n", addrName.c_str(), howmuch);
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


mitchellcash:
This is just changing and adding printfs so LGTM. Although it wouldn't hurt to get this tested. 

What's the best way to get a misbehaving node? If I run a lower protocol version is that my node misbehaving? Second, we have this code:
```
// Do not ban addresses on the testnet for misbehaving
if (fTestNet)
    return false;
```
Does this mean we can't test this on the testnet or does it mean we should still get logs for misbehaving nodes it just won't ban them?

IngCr3at1on:
@mitchellcash I put that check in there cause I was annoyed at being banned when trying to test microprimes...

Probably the easiest way to test it would be to compile a wallet that testnet check AND the connect input related microprime checks commented out. Deploy that on an isolated network then take a second (non-modified wallet) and run a microprime in it.

IngCr3at1on:
Being that this is only print messages I think it's safe to merge; could be handy in debugging if the current fork issue happens again and if there is an issue with the messages it's easy enough to correct.

PR #220, opened 2015-09-21
= instead of == in multisig_tests.cpp

jwrb:
Taken from [1496](https://github.com/bitcoin/bitcoin/pull/1496)


IngCr3at1on:
Seems sane and the tests pass (though they already did...)

They won't build without #221 of course.

PR #225, opened 2015-09-22
RPC: Forbid RPC username == RPC password

jwrb:
Added security measure taken from [#1987](https://github.com/bitcoin/bitcoin/pull/1987)

Compiles & Runs as expected


mitchellcash:
Me being lazy when I've tested Paycoin in the past I've used:
```
rpcuser=test
rpcpassword=test
```

Guess I'm going to have to change that :P

jwrb:
Yeah, security though. Some servers out there could be using this so it's best to at least try and secure them.

mitchellcash:
Yeah agreed, I'm all for security improvements and every bit counts, LGTM.

IngCr3at1on:
I agree with @mitchellcash LGTM

PR #226, opened 2015-09-23
CHashWriter that does SHA256 in-place during serialization

IngCr3at1on:
I confirmed this builds in gitian and used it to sync a chain as well as left it running on a testnet prime overnight, everything appears to be functioning properly (if anyone wants to test that's great of course).

I have not been able to test CPU load in comparison to without this function but based on the authors notes it seems viable.

Reference: https://github.com/bitcoin/bitcoin/pull/1567

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -611,6 +611,46 @@ inline uint256 Hash(const T1 pbegin, const T1 pend)
     return hash2;
 }
 
+class CHashWriter
+{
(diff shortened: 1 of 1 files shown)


mitchellcash:
utACK

mitchellcash:
Tested, compiles and runs nicely through an IBD.

PR #227, opened 2015-09-23
Update Icons

jwrb:
Update the tray icons to show the correct testnet and toolbar icons

This compiles and runs as expected


mitchellcash:
@jwrb can you please make the testnet icon this colour: https://github.com/PaycoinFoundation/paycoin/blob/master/src/qt/res/images/splash_testnet.png

Because we already have a chosen testnet colour in the form of the splash screen best to keep it the same and if necessary have a seperate conversation about colour changes.

IngCr3at1on:
This requires rebase

mitchellcash:
Outside of having this rebased before merge we should discuss the testnet icon colour. We need to pick a colour path for testnet and stick to it.

jwrb:
This would be a wise idea. Perhaps something to bring up with the team? 
This is
a design choice, and although not seen by many it will still remain present 
in
every single release.

IngCr3at1on:
I agree in regards to more discussion regarding what to actually change/modify the icons to however if the tooltip icon is not the testnet icon I would suggest submitting the code patch in a separate PR.

mitchellcash:
Does anyone have any further thoughts on colour schemes? I'm no designer but using a Complimentary Colour Tool I found, I put the base colour layer of `0395B8` into the tool and it spat out some possible complimentary colours of `FF7F00`, `6E07C2` or `FFF500`.

http://paletton.com/#uid=73n1f0kIGDCrlvJxMBOOCAzTyyp

Seemed like just as good of a way to pick a testnet colour than any other way haha!

Thoughts?

IngCr3at1on:
Yeah I don't know, I'm generally not the best to ask about that stuff lol

mitchellcash:
Not me either but if I was to make a pick it would be `FF7F00`

PR #233, opened 2015-09-25
Encapsulate BDB environment inside new CDBEnv class

IngCr3at1on:
Minor differences between this and the commits in question (specifically cd9696f) as this doesn't account for the needed changes to compile QT; was unable to locate a patch for this so instead wrote in the required functionality myself (subsequently this may not exactly match bitcoins implementation, I believe they actually did something similar at a later time (based on the commits where BDB was removed from everything but the wallet).

This (among other things) is needed to incorporate salvage functionality for corrupt wallet.dat files.

Works fine in my testing on linux but should definitely have some extended testing in all supported OS.

reference: https://github.com/bitcoin/bitcoin/pull/1293

### DIFF
--- a/src/db.cpp
+++ b/src/db.cpp
@@ -145,10 +145,7 @@ CDB::CDB(const char *pszFile, const char* pszMode) :
             {
                 delete pdb;
                 pdb = NULL;
-                {
-                     LOCK(bitdb.cs_db);
-                    --bitdb.mapFileUseCount[strFile];
-                }
+                --bitdb.mapFileUseCount[strFile];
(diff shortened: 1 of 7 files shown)


IngCr3at1on:
Added https://github.com/bitcoin/bitcoin/pull/1728 because I see little reason not to.

IngCr3at1on:
This is required for mock db capabilities which are needed for unit tests.

IngCr3at1on:
Has anyone looked at this yet (@PaycoinMaster, @mitchellcash)?

PR #234, opened 2015-09-25
Remove CTxDB::ReadOwnerTxes.

IngCr3at1on:
It seems it wasn't ever used since 0.1.5.


IngCr3at1on:
Added the trivial label cause this code is unused anyway.

PR #235, opened 2015-09-27
clear path cache after getting a proper config file

mitchellcash:
Fixes the below:

1. Create an empty data directory with a testnet paycoin.conf; e.g.
    mkdir /tmp/d
    echo "testnet=1" >> /tmp/d/paycoin.conf
    echo "rpcuser=abc" >> /tmp/d/paycoin.conf
    echo "rpcpassword=def" >> /tmp/d/paycoin.conf

2. Run Paycoin-Qt with -datadir=/tmp/d

You get a nice orange splashscreen, BUT a wallet.dat and blockchain files directly in /tmp/d, not in /tmp/d/testnet3/

paycoind behaves correctly.

Bitcoin reference:
https://github.com/bitcoin/bitcoin/commit/b357a71cfaf99dd2e7d203a12edfcc5df4a93386

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -79,6 +79,7 @@ bool fNoListen = false;
 bool fLogTimestamps = false;
 CMedianFilter<int64> vTimeOffsets(200,0);
 bool fReopenDebugLog = false;
+bool fCachedPath[2] = {false, false};
 
 // Init openssl library multithreading support
 static boost::interprocess::interprocess_mutex** ppmutexOpenSSL;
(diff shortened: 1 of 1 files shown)


mitchellcash:
Label: `Bug`

IngCr3at1on:
This doesn't occur in Linux, just for knowledge sake do you know if it happens on Mac or Windows or both?

mitchellcash:
Hmm I'm embarrassed to admit I didn't actually test this one thinking it was a safe commit. I also can't reproduce this on Windows or Mac.

I still think it should be merged as it is still following the same code path as what Bitcoin did but it has no effect so up to everyone.

IngCr3at1on:
Hahaha... glad I'm not insane :P

It has no ill effect so far as I can tell so I've got no problem with it, changing the label though lol

PR #236, opened 2015-09-27
fix crash: CoinControl "space" bug

mitchellcash:
Bug description:
1. Run Paycoin-Qt
2. Open CoinControl
3. Click mouse here(Don't select any input)

4. Press "space"
5. Result

Fix:
this->currentItem() returns the current item in the tree widget and NULL if no selected items.
So we don't try setCheckState for NULL.


mitchellcash:
Labels: `GUI`, `Bug`

IngCr3at1on:
Confirmed issue and fix, builds in gitian. LGTM

PR #238, opened 2015-09-30
Check primenode keys on app start

IngCr3at1on:
Previously entering an invalid primenode key would start the wallet without error, if it was an invalid format it had the potential to throw an error otherwise the wallet would attempt to stake and not say the key was invalid until finding and creating a stake transaction and attempting to Accept the block (failing during ConnectInputs).

This will create a test signature and process it during init prompting if the key is invalid.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -156,15 +156,25 @@ bool AppInit(int argc, char* argv[])
     return fRet;
 }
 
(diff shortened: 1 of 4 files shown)


IngCr3at1on:
Requires #237 (will require rebase after #237's merge).

PR #239, opened 2015-10-01
Script to be able to perform copyright header updates

mitchellcash:
Referenced from https://github.com/bitcoin/bitcoin/commit/57702541a2009d4b936ebfddeeec3db7f8f74e5c

Original author @gubatron

This is being put here for future use as no current files were modified when using the script.

### DIFF
--- a/contrib/devtools/README.md
+++ b/contrib/devtools/README.md
@@ -1,0 +1,11 @@
+## fix-copyright-headers.py
+
+Every year newly updated files need to have its copyright headers updated to reflect the current year.
+If you run this script from src/ it will automatically update the year on the copyright header for all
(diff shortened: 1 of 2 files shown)


gubatron:
ha, good to know this is in use :) cheers.

IngCr3at1on:
Interesting

IngCr3at1on:
The real question on this is if anyone will be using the script enough for it to need to be present.

mitchellcash:
It will come in handy come 2016 and seeing as it's going into the devtools folder and not burdening the actual Paycoin core codebase I see no harm in leaving it in devtools for possible future use. It's a useful script and if needed come 2016 I'll make sure to use it ;)

PR #245, opened 2015-10-02
Use a CriticalSection when writing to the primenode database

IngCr3at1on:
The odds of multiple threads ever attempting to write to this database
when it's only writeable when being created are pretty much impossible
but best practice is to lock the database anyway so we will.


IngCr3at1on:
This has been rebased over master do to conflict in primenodes.h

ligerzero459:
LGTM

PR #247, opened 2015-10-02
Fix display of external scrape transactions in qt wallet.

IngCr3at1on:
There's a weird issue in the QT where it doesn't always calculate the correct credit value on external stakes (sometimes and often resulting in displaying the complete stake value). The scrape value is always the value of the last output of the transaction so check it when setting the amount for an external scrape.

Label external scrapes differently then standard stakes and always list the address scraped from.

Before:
selection_135

After (with an address label set in address book):
selection_134

### DIFF
--- a/src/qt/transactionrecord.cpp
+++ b/src/qt/transactionrecord.cpp
@@ -49,8 +49,15 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
             CTxDestination address;
             CTxOut txout = wtx.vout[1];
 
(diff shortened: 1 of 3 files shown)


mitchellcash:
Argh, I tested this a month ago but I forgot to comment. I'm almost certain this fixed things. I will have to test again maybe.

IngCr3at1on:
Got @krumblez to test this to confirm scrape transactions looked right. But I think he said both the scrape address and staking address were in the same wallet so unfortunately doesn't quite confirm it.

IngCr3at1on:
@krumblez has confirmed this is working properly.

Issue #241, opened 2015-10-01
Error with printcoinage debug arg


IngCr3at1on:
Use of the printcoinage arg for extra debugging causes random and silent daemon crashes.

IngCr3at1on:
Fix merged, closing

IngCr3at1on:
Mixed up commits, this was not merged yet...

IngCr3at1on:
Fix merged, closing issue

PR #250, opened 2015-10-05
Fix issue #241  where -printcoinage causes silent crashes in linux

IngCr3at1on:


IngCr3at1on:
@mitchellcash does this effect mingw compiling in windows at all (assuming not I want to go ahead and merge it as it's fairly minor)?

mitchellcash:
This compiles fine on mingw. Let's get it merged

PR #251, opened 2015-10-05
Replace instances of lld/llu with the PRI64d/PRI64u macro for MSVC++ compat

IngCr3at1on:


IngCr3at1on:
I'm sure there are other things that will stop MSVC++ from compiling but it's best to use the defined macros in this instance either way.

IngCr3at1on:
This has been rebased over master. I'd like to confirm it doesn't somehow effect mingw compiling (@mitchellcash)  before merging and I'm still curious if it helps with MSVC++ compiling (though I expect many other errors will occur).

mitchellcash:
mingw compiling is fine :) can't comment on MSVC++

IngCr3at1on:
@mitchellcash someone surely will test MSVC++ eventually, either way I'm in favor of using the macro instead of lld. Just going to double check if I missed any instances of it lol.

PR #254, opened 2015-10-06
Run BDB disk-less for test_paycoin

IngCr3at1on:
Reference: https://github.com/bitcoin/bitcoin/pull/1246

PR #256, opened 2015-10-06
Very basic accounting tests for general sanity checking

IngCr3at1on:
Will be replacing these later after updating the accounting api


IngCr3at1on:
Requires #254

IngCr3at1on:
Rebased over master and replaced some instances of BOOST_CHECK w/ BOOST_CHECK_EQUAL (where viable).

Issue #17, opened 2014-12-23
Minting tab updates


paulc010:
Just a cosmetic issue but when outputs are spent or staked they aren't being removed from the minting tab, although new outputs are appearing correctly. Restarting the wallet clears the issue.

Example: If I have 1 output and it stakes and splits, then the minting tab now shows 3 entries - the original "new staked output" (incorrect) plus the two resulting split outputs from the staking tx.

IngCr3at1on:
This should have been resolved with #262, closing for now. Please feel free to re-open with more details if there are further issues.

PR #262, opened 2015-10-08
Remove spent transactions from the minting table model.

IngCr3at1on:
This addresses #17


mitchellcash:
Tested and confirmed working. I like it, good to see the minting table cleaned up a bit on a always on Qt wallet.

PR #267, opened 2015-10-15
Bugfix messagestart

IngCr3at1on:
Unfortunately I've all but ruled this out as the cause of the fork issues:
NONE of the debug logs I've looked at from a forked node have the error that would show if this was causing an issue and as it stands the "bad" start message isn't being sent anywhere but needs to be corrected anyway for obvious reasons.

### DIFF
--- a/src/db.cpp
+++ b/src/db.cpp
@@ -11,6 +11,7 @@
 #include "util.h"
 #include "main.h"
 #include "kernel.h"
+#include "protocol.h"
 #include <boost/version.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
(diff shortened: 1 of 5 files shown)


mitchellcash:
Why did we need the below added to both CAddrDB::Write and CAddrDB::Read?

```
+    unsigned char pchMessageStart[4];
+    GetMessageStart(pchMessageStart);
```

Shouldn't adding the protocol.h include be enough?

Also what's the reason for storing pchMessageStart in protocol.cpp. It just seems an unnecessary diversion from Bitcoin.

And lastly, were we sure that the removal of pchMessageStart from main.cpp (even though it was using a different network magic number) didn't have any other negative effects?

Sorry somehow I landed on these commits when I was looking at other things and was just curious :)

IngCr3at1on:
pchMessageStart is a blank character array, it's being initialized so GetMessageStart can be called on it (so yes that's valid).

Frankly I'm not sure I'm concerned with whether it diverges from bitcoin or not. If we get around to creating/doing a Chain Parameters class it will likely be moved anyway.

The possible negative effects that you are talking about would have been fair more likely to occur when  I had added it there inadvertently and it was invalid (something no-one else caught previously).

mitchellcash:
Thanks for the info. I agree with the move to Chain Parameters when that happens.

PR #268, opened 2015-10-15
Add a '-stake' flag to disable the staking thread

IngCr3at1on:
Default value is 1 (staking enabled)
When set to 0 either by command line or conf file the staking thread is never started.

This is handy for machines being used purely as dnsseeds or block explorer wallets to
reduce the CPU overhead of the wallet itself.


IngCr3at1on:
Fixed commit message where I wrote "-staking" instead of "-stake"...

PR #270, opened 2015-10-16
Add the testnet genesis block to the testnet checkpoints

IngCr3at1on:
Also changes 546 to use hex format like normal checkpoints (0x...)


mitchellcash:
LGTM

PR #273, opened 2015-10-17
Upgrade OpenSSL to 1.0.1h

mitchellcash:
Upgrade for https://www.openssl.org/news/secadv/20140605.txt

First: there is no vulnerability that affects ecdsa signing or verification. However, the MITM attack vulnerability (CVE-2014-0224) may have some effect on our usage of SSL/TLS.

As long as payment requests are signed (which is the common case), usage of the payment protocol should also not be affected.

The TLS usage in RPC may be at risk for MITM attacks. If you have -rpcssl enabled, be sure to update OpenSSL as soon as possible.

Reference:
https://github.com/bitcoin/bitcoin/commit/6e7c4d17d8abb4b1c8b91504699ce6970e01a1fb

### DIFF
--- a/contrib/gitian-descriptors/deps-linux.yml
+++ b/contrib/gitian-descriptors/deps-linux.yml
@@ -16,7 +16,7 @@ packages:
 reference_datetime: "2013-06-01 00:00:00"
 remotes: []
 files:
-- "openssl-1.0.1g.tar.gz"
+- "openssl-1.0.1h.tar.gz"
 - "miniupnpc-1.9.tar.gz"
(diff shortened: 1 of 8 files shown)


mitchellcash:
@IngCr3at1on could you please test/review this one?

IngCr3at1on:
Yeah I'll throw that in a gitian build after the one I have going is done. Assuming it builds I don't foresee any issues but it would be nice to get someone to test the windows and mac binaries (once they're done).

IngCr3at1on:
Gitian builds are fine, looks good.

PR #274, opened 2015-10-17
Move most explicit getters in optionsmodel to header

mitchellcash:
- is more consistent and saves quite some lines of code


IngCr3at1on:
LGTM

PR #276, opened 2015-10-17
Bugfix: remove conflicting transactions from memory pool

mitchellcash:
When a transaction A is in the memory pool, while a transaction B (which shares an input with A) gets accepted into a block, A was kept forever in the memory pool.

This commit adds a CTxMemPool::removeConflicts method, which removes transactions that conflict with a given transaction, and all their children.

This results in less transactions in the memory pool, and faster construction of new blocks.

Reference:
https://github.com/bitcoin/bitcoin/commit/231b399952fd620ee0f72b1947024dba9651630d

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -721,7 +721,7 @@ bool CTxMemPool::addUnchecked(const uint256& hash, CTransaction &tx)
 }
 
 
-bool CTxMemPool::remove(CTransaction &tx)
+bool CTxMemPool::remove(const CTransaction &tx, bool fRecursive)
 {
     // Remove transaction from memory pool
     {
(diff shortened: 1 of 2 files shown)


mitchellcash:
I will need someone to assist with some testing on this though.

IngCr3at1on:
We should be able to test this by just leaving it running for a bit and allowing reorganizations to occur on their own. Beyond that some sort of unit tests for the mempool/transactions would be totally awesome.

mitchellcash:
Rebased over master (plus I wanted Travis to run his eyes over it ;) )

IngCr3at1on:
This has been running for several days on a staking node (on mainnet) without issues. Memory footprint does appear to be reduced as well when compared to other (non-paycoin) nodes on the same machine.

mitchellcash:
That's what I like to hear :)

PR #277, opened 2015-10-18
Qt: small header changes / fixes

mitchellcash:
- ensure header inclusion guard is named after the header file
- add missing comments at the end of some inclusion guards

Reference:
https://github.com/bitcoin/bitcoin/commit/128eefa0f88ff80c5d6398d13e25a3f4df96081a

### DIFF
--- a/src/qt/qtipcserver.h
+++ b/src/qt/qtipcserver.h
@@ -1,4 +1,10 @@
+#ifndef QTIPCSERVER_H
+#define QTIPCSERVER_H
+
+// Define Bitcoin-Qt message queue name
 #define BITCOINURI_QUEUE_NAME "BitcoinURI"
 
 void ipcInit();
 void ipcShutdown();
+
+#endif // QTIPCSERVER_H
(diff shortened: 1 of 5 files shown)


IngCr3at1on:
You missed qtipcserver.h

mitchellcash:
@IngCr3at1on okay fixed but can you please confirm looks okay

src/qt/qtipcserver.h
```diff
@@ -1,4 +1,10 @@
+#ifndef QTIPCSERVER_H
+#define QTIPCSERVER_H
+
+// Define Bitcoin-Qt message queue name
 #define BITCOINURI_QUEUE_NAME "BitcoinURI"
```
IngCr3at1on: I'd expect this to be PaycoinURI but being that the URI stuff is in a separate branch you probably have already changed that there.

IngCr3at1on:
Looks good with one style issue, IMHO all these files should end with a new line (I notice this was removed in one case but otherwise most of them are lacking it in the first place so maybe this isn't the commit to address that in).

mitchellcash:
@IngCr3at1on they all do have a new line at the end of them ;) and the one I removed was me actually removing one of two trailing new lines. GitHub in the web displays them weird so it's best to open them up in a text editor.

PR #278, opened 2015-10-18
Simplify CDBEnv::Open() / fix small glitches

mitchellcash:
- remove pathEnv from CDBEnv, as this attribute is not needed
- change path parameter in ::Open() to a reference
- make nDbCache variable an unsigned integer

Reference:
https://github.com/bitcoin/bitcoin/commit/c74bae0fdf211551e5a45d1b6bbc766d91381ad9

### DIFF
--- a/src/db.cpp
+++ b/src/db.cpp
@@ -62,7 +62,7 @@ void CDBEnv::Close()
     EnvShutdown();
 }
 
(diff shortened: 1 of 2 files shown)


IngCr3at1on:
LGTM

PR #283, opened 2015-10-21
Move help messages out of AppInit2

IngCr3at1on:
- Removes \t's from help message that are removed afterwards anyway
- Moves UI-specific command-line options help to UI code
- Moves "-detachdb" out of #ifdef USE_UPNP

PR #284, opened 2015-10-21
Reorder AppInit2

IngCr3at1on:
This reorders AppInit2 into 11 sections:
1) basic setup (OS stuff, signal handlers)
2) parameter interactions (softset)
3) interpreting options without side effects (just setting internal variables)
4) application initialization (datadir lock, daemonize, pidfile, log file)
5) network intialization (network params, proxy, binding)
6) load blockchain
7) load wallet
8) load peers
9) load primenodes and scrape databases
10) start node/rpc threads
11) finished

This means invalid configurations will be detected as soon as possible
(in partciular, before loading the block chain).

reference: bitcoin/bitcoin#1369

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -118,10 +118,6 @@ void HandleSIGHUP(int)
     fReopenDebugLog = true;
 }
 
-
-
-
-
 //////////////////////////////////////////////////////////////////////////////
 //
 // Start
@@ -300,6 +296,7 @@ std::string HelpMessage()
 
 bool AppInit2()
 {
+    // ********************************************************* Step 1: setup
 #ifdef _MSC_VER
     // Turn off Microsoft heap dump noise
     _CrtSetReportMode(_CRT_WARN, _CRTDBG_MODE_FILE);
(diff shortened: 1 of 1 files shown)


IngCr3at1on:
This includes #283 cause otherwise I couldn't push it, will require a minor rebase after merging #283

IngCr3at1on:
This has been rebased over master.

IngCr3at1on:
This has been rebased over master once again. It should be tested and checked carefully to make sure I didn't inadvertently revert something.

IngCr3at1on:
Updated commit message to say 11 sections not 10.

mitchellcash:
What can I do to help test this, I'd like to get this in the codebase?

IngCr3at1on:
@mitchellcash everything is related to init itself so really the fact that it starts properly is the main thing. You could test some of the init functions like connect, etc but there's little reason they wouldn't work the same as they did before.

There was an issue I was having getting rpc to work on a testnet daemon properly at one point but that's cause the check for it was happening too late.

The main point of this is to detect configuration errors before actually starting anything instead of having everything mixed in at weird points the way it was before, forcing you to go "where should I put this?".

src/init.cpp
```diff
@@ -362,6 +398,38 @@ bool AppInit2()
...
+        if (!ParseMoney(mapArgs["-paytxfee"], nTransactionFee) || nTransactionFee < MIN_TX_FEE)
+            return InitError(strprintf(_("Invalid amount for -paytxfee=<amount>: '%s'"), mapArgs["-paytxfee"].c_str()));
+
+        if (nTransactionFee > 0.25 * COIN)
+            InitWarning(_("Warning: -paytxfee is set very high.  This is the transaction fee you will pay if you send a transaction."));
+    }
+
+    // ********************************************************* Step 4: application initialization: dir lock, daemonize, pidfile, debug log
+
+    // Make sure only a single bitcoin process is using the data directory.
```
mitchellcash: I know you mentioned a more widespread sweeping change of the leftover bitcoin, peercoin naming in comments to paycoin but seeing as this is part of an existing pull request with different motives can you please change this comment to refer to paycoin instead of bitcoin.

mitchellcash:
Apart from the comment change I've tested this and it looks good to me.

IngCr3at1on:
@mitchellcash fixed the comment

mitchellcash:
@IngCr3at1on thanks and now LGTM

PR #293, opened 2015-10-23
Don't stop users from using 0.0.0.0 as the a bind address

IngCr3at1on:
While 0.0.0.0 is already the default bind address for IPv4 blocking
it breaks old conf files that are configured using this.

Instead of removing the IsValid check on bind addresses don't check
if the address is a 0 address during init (allow other IsValid checks
to function as they always have).

0, 0.0 and 0.0.0 addresses are all expanded to 0.0.0.0 (confirmed by
printing to string during tests) while anything longer is obviously
not an IPv4 address and should be checked like normal.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -695,7 +695,7 @@ bool AppInit2(int argc, char* argv[])
         if (mapArgs.count("-bind")) {
             BOOST_FOREACH(std::string strBind, mapMultiArgs["-bind"]) {
                 CService addrBind(strBind, GetListenPort(), false);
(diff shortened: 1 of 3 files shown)


IngCr3at1on:
This should fix @xmk3's complaint regarding the IsValid check (added when cleaning up InitErrors) stopping the usage of 0.0.0.0 addresses as bind addresses.

jwrb:
I'll test this shortly.

jwrb:
This works as expected. LGTM :+1:

PR #315, opened 2015-10-28
Fix regression in tests introduced in ab3a8eb2e9f715f137f9a4077de71df…

IngCr3at1on:
…f56c27d73

### DIFF
--- a/src/primenodes.cpp
+++ b/src/primenodes.cpp
@@ -188,10 +188,10 @@ bool CTransaction::IsPrimeStake(CScript scriptPubKeyType, CScript scriptPubKeyAd
     }
 
     // Confirm the stake passes the minimum for a primenode
-    if (nTime >= END_PRIME_PHASE_ONE && GetValueOut() < MINIMUM_FOR_PRIMENODE_PHASE2)
-        return DoS(100, error("IsPrimeStake() : credit doesn't meet requirement for primenode = %lld while you only have %lld", MINIMUM_FOR_PRIMENODE_PHASE2, GetValueOut()));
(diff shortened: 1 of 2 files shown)


src/primenodes.cpp
```diff
@@ -188,10 +188,10 @@ bool CTransaction::IsPrimeStake(CScript scriptPubKeyType, CScript scriptPubKeyAd
     }
 
     // Confirm the stake passes the minimum for a primenode
-    if (nTime >= END_PRIME_PHASE_ONE && GetValueOut() < MINIMUM_FOR_PRIMENODE_PHASE2)
-        return DoS(100, error("IsPrimeStake() : credit doesn't meet requirement for primenode = %lld while you only have %lld", MINIMUM_FOR_PRIMENODE_PHASE2, GetValueOut()));
-    if (GetValueOut() < MINIMUM_FOR_PRIMENODE_PHASE1)
-        return DoS(100, error("IsPrimeStake() : credit doesn't meet requirement for primenode = %lld while you only have %lld", MINIMUM_FOR_PRIMENODE_PHASE1, GetValueOut()));
+    if (nTime >= END_PRIME_PHASE_ONE && nValueOut < MINIMUM_FOR_PRIMENODE_PHASE2)
```
IngCr3at1on: This was originally changed from GetValueOut() to nValueOut (which is assigned in main.cpp during ConnectInputs() using the normal GetValueOut()) to allow for tests to work in the first place.

IngCr3at1on:
This appears to have been missed during a rebase and has only effected tests.

mitchellcash:
I'm mucking around with tests at the moment and would love this merged!

Confirmed that this patch fixes the regression in primenode tests.

PR #321, opened 2015-11-02
Update Hard Coded DNS IP address

HyperCommNet:
Replaced old 32bit ip addresses with more current addresses.


jwrb:
This is messy. This only needs to be a single commit.

HyperCommNet:
Hey i am learning here...

jwrb:
Landed in #327

PR #324, opened 2015-11-02
Remove extra debugging on accounting tests.

IngCr3at1on:
Now that they're working I don't think we need this running constantly.


mitchellcash:
LGTM

Issue #312, opened 2015-10-28
add link from console to wiki to "define help commands"


mdpfeiffer:
Since the wiki increasingly explains the function of the various commands available via help > debug > console, users less familiar with code might appreciate a command that links them to online explanations.

IngCr3at1on:
This is resolved with #326

PR #326, opened 2015-11-03
Include Link To Wiki

jwrb:
Include a link on the Debug console page to the Wiki, allowing for access to more detailed information about possible commands & their usage.

This should solve [#312](https://github.com/PaycoinFoundation/paycoin/issues/312)


IngCr3at1on:
This works on my linux box and in gitian builds. Relatively trivial otherwise so I'm merging it.

PR #327, opened 2015-11-03
Update Hard coded DNS IP

jwrb:
This replaces #321. Cleaned extra commits out

Update the hard coded DNS seeds to point to Paycoin, the Paycoin Foundation and the Paycoin core development group.

Thanks @HyperCommNet


jwrb:
After digging, it seems that the DNS seeds are actually incorrect @HyperCommNet

scroogemcdev:
As I understand it these are only ever used in the event DNS resolution isn't available (which rarely occurs in an instance where the nodes would still work anyway). Merging as I'm not sure we have any idea what those original nodes are...

PR #330, opened 2015-11-03
Better validity checks for scrape addresses

IngCr3at1on:
Check IsValid on all address instances instead of jumping to IsMine;
	provides clearer error information

Return error codes for scrape addresses;
	used -1 for most, is for misc error

Check if a wallet is encrypted and if it's locked when making changes
to the scrapes database.

Add full tests for rpc scrape functionality.

### DIFF
--- a/src/bitcoinrpc.h
+++ b/src/bitcoinrpc.h
@@ -26,6 +26,9 @@ json_spirit::Object JSONRPCError(int code, const std::string& message);
 /** Convert parameter values for RPC call from strings to command-specific JSON objects. */
 json_spirit::Array RPCConvertValues(const std::string &strMethod, const std::vector<std::string> &strParams);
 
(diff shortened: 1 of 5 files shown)


IngCr3at1on:
This is a bug because currently a user can modify a scrape address on a locked wallet without authentication.

This requires #315 for the tests to run.

IngCr3at1on:
This has been rebased over master and added a commit to remove a now-unneeded extern (can be squashed after review).

mitchellcash:
Confirmed testing in both the Qt and also paycoind with RPC calls. Everything looks good to me.

scroogemcdev:
Rebased over master to accommodate #421

IngCr3at1on:
Rebased over master to accommodate #421

PR #331, opened 2015-11-03
Enables log timestamps by default

jwrb:
Thanks to @arcanis for this one.

Log timestamps by default when logging. This can be handy for debugging and finding out when things went wrong. (using the argument can obviously turn this off)

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -358,7 +358,7 @@ bool AppInit2(int argc, char* argv[])
 #endif
     fPrintToConsole = GetBoolArg("-printtoconsole");
     fPrintToDebugger = GetBoolArg("-printtodebugger");
-    fLogTimestamps = GetBoolArg("-logtimestamps");
+    fLogTimestamps = GetBoolArg("-logtimestamps", true);
 
 #ifndef QT_GUI
     for (int i = 1; i < argc; i++)


mitchellcash:
Confirmed this works

But do we want this on by default? Logfile space, possible privacy concerns. I like it being off be default but it doesn't bother me too much.

IngCr3at1on:
Log space was my concern. I'm wondering if the compromise should be to only enable timestamps by default if debug is enabled.

ligerzero459:
That sounds like a fair trade off. Timestamps wouldn't be as big of a deal if you're in debug mode. You expect large log files

mitchellcash:
If we do go ahead with this (which I'm in favour of) it might be a good idea to also change:
```
"  -logtimestamps   \t  " + _("Prepend debug output with timestamp") + "\n" +
```
to
```
"  -logtimestamps   \t  " + _("Prepend debug output with timestamp (default: 1)") + "\n" +
```

scroogemcdev:
Definitely want to update the help message if doing this yes. Doesn't seem like we have a consensus though... @xmk3 or @HyperCommNet care to comment?

CeForce:
Makes sense to have timestamps by default when debug is on to me. Seems like a fair trade to me as well. If you're debugging, odds are you want those time stamps anyways. :+1:

IngCr3at1on:
Closing in favor of #479

PR #332, opened 2015-11-04
Add .travis.yml to run all linux gcc builds and tests

IngCr3at1on:


mitchellcash:
Nice, got it in. Now to just get the clang builds working ;)

PR #335, opened 2015-11-05
Cleanup checkpoints

mitchellcash:
After discussion with @IngCr3at1on I think this will need some review and testing to understand the full effect of updating the checkpoints like I have.


IngCr3at1on:
I synced a node overnight using these checkpoints without issues. This is safe enough.

Issue #294, opened 2015-10-24
Syncing stalled


RdeWilde:
I had two nodes syncing thru 'connect' to two nodes with blockchain validated against ledger.

Now both nodes are stuck:
http://108.61.103.237/ stuck at 328798 328798 https://ledger.paycoin.com/block/84f73630aafe8f3e371bdbe12a517377092332314aba184155eb4edc6829cffa
http://104.238.187.227/ stuck at 317886 https://ledger.paycoin.com/block/4cfb3104b94cd7bd006203e1a7d99a570c639d2a0877159e5f6756b83f4fd73c

In the debug log there are many askfor block lines and also many of these errors:

```ERROR: CheckProofOfStake() : INFO: read txPrev failed
WARNING: ProcessBlock(): check proof-of-stake failed for block 151eb6b86abbd3cf79b020b6f81e5aa1027092c9cabfd182fadc9faf9ca8b977
received block 25407f5c465bc0c0b45c```

```ERROR: CheckProofOfStake() : INFO: check kernel failed on coinstake a6a95922fe324f4089dac4dbe33e8068e1237a6f36a05d9bd6d3214d406f59fc, hashProof=0000000000000000000000000000000000000000000000000000000000000000
WARNING: ProcessBlock(): check proof-of-stake failed for block 0753fffdd93196cd9aee163d51741884e9002fcca7f9260b3eccd63b5b2b21cb
received block 6b7b91378026b2edd430
ERROR: CheckProofOfStake() : INFO: read txPrev failed```

Also stuff like this (which is normal I think)
askfor block 16649189eb80f9d0d1f7   1445726500000000
askfor block 89f59de7e1ae1225ebe4   1445726500000000
askfor block 15444ef9336bd839fe8a   1445726380000000
askfor block 0c9c2149fa61ceceb5eb   1445726380000000
askfor block f0033f21df779fd97938   1445726380000000
askfor block 2986ffc0855684c8e1c1   1445726260000000
askfor block 482c9d0207d312e3f814   1445725420000000
askfor block e24248e239055d146595   1445725300000000
askfor block cf99d0037473f9fda792   1445725180000000
askfor block a5bcb24e13186363d0a0   1445723500000000
askfor block 6eecf10d2004daf23cfc   1445723500000000
askfor block 5c18f57353d75b1b31a3   1445723500000000
askfor block 9068620338cbeb87ffc6   1445726620000000
askfor block f4696c9652bca4c0f833   1445726260000000
[...]

PR #337, opened 2015-11-05
Fixed initial download stall

ligerzero459:
Adapted from Peershares/Peershares#110. Peercoin solved the download stall by using the latest PoW block as reference to download, but because Paycoin only uses PoS that doesn't work.

Peershares solved this by making the reference block the best block, no matter the proof.

Block requests are also retried immediately during initial download because the same last block is asked multiple times until we have the full chain.

This should fix issue #294

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2178,6 +2178,11 @@ bool ProcessBlock(CNode* pfrom, CBlock* pblock)
         if (!CheckProofOfStake(pblock->vtx[1], pblock->nBits, hashProofOfStake))
         {
             printf("WARNING: ProcessBlock(): check proof-of-stake failed for block %s\n", hash.ToString().c_str());
+
+            // paycoin: ask for missing blocks
+            if (pfrom)
+                pfrom->PushGetBlocks(pindexBest, pblock->GetHash());
+
             return false; // do not error here as we expect this during initial block download
         }
         if (!mapProofOfStake.count(hash)) // add to mapProofOfStake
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -3046,7 +3051,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv)
                         // download node to accept as orphan (proof-of-stake
                         // block might be rejected by stake connection check)
                         vector<CInv> vInv;
-                        vInv.push_back(CInv(MSG_BLOCK, GetLastBlockIndex(pindexBest, false)->GetBlockHash()));
```
IngCr3at1on: Any chance at updating the comments above this ;)

src/net.h
```diff
@@ -305,7 +305,10 @@ class CNode
         nLastTime = nNow;
 
         // Each retry is 2 minutes after the last
```
IngCr3at1on: Also the comment here could use a minor update iirc.

src/main.cpp
```diff
@@ -2178,6 +2178,11 @@ bool ProcessBlock(CNode* pfrom, CBlock* pblock)
         if (!CheckProofOfStake(pblock->vtx[1], pblock->nBits, hashProofOfStake))
         {
             printf("WARNING: ProcessBlock(): check proof-of-stake failed for block %s\n", hash.ToString().c_str());
+
+            // peershares: ask for missing blocks
```
mitchellcash: Also please update this comment :)

IngCr3at1on: Why? The patch it from peershares.

ligerzero459: That was my reasoning as well, but I can update it :smile:

PR #338, opened 2015-11-05
Sync fix for 0.3.2.0

ligerzero459:
0.3.2.0 nodes refuse to sync properly past block 328798 because of a duplicate key in the Prime Keys database. Block 328799 occurs in the 14 week window, but the duplicate key means it's not valid until 15 weeks, causing the node to reject all blocks with that key, preventing it from syncing.

This PR removes the second duplicate, ensuring that the key is valid at the timestamp on block 328799.


RdeWilde:
Nice1

PR #339, opened 2015-11-05
Flag primenode database to be updated

IngCr3at1on:
This goes specifically with #338 but we'll be using the database version more in the future.


ligerzero459:
:+1: LGTM

PR #342, opened 2015-11-05
Minor update to protocol (not minimum protocol, no major effect)

IngCr3at1on:
This is just so that the alert I sent out on the testnet gets
cancelled, otherwise they're stuck with an annoying message until
the next update.


IngCr3at1on:
Since we're doing a new release in the relatively near future this really isn't needed.

PR #353, opened 2015-11-08
Fix comment in version.h

mitchellcash:


IngCr3at1on:
When the original devs did find and replace they replaced every "peercoin" comment with "paycoin", I have had no motivation to correct this. I'm not really sure I see a point in fixing one comment when there are countless others that are still wrong.

mitchellcash:
Yeah I know, I too was being lazy. I just happened to be looking at version.h. Leave it with me for a while I'll get some more name fixes into this commit.

mitchellcash:
Closed in favour of #393.

PR #356, opened 2015-11-08
add Alt-Shortcut for "Clear All" in sendcoinsdialog.ui

mitchellcash:


IngCr3at1on:
None of these work right on Arch for whatever reason so I'll have to confirm this one on my laptop at some point.

mitchellcash:
It's strange that on Arch but Ubuntu and Windows should test fine according to my tests :P

PR #358, opened 2015-11-09
Make it easier to read/copy ubuntu commands

RdeWilde:


doc/build-unix.md
```diff
@@ -61,13 +61,17 @@ Versions used in this release:
...
+- sudo apt-get install libssl-dev
+- sudo apt-get install libdb4.8-dev
+- sudo apt-get install libdb4.8++-dev
+
+Boost 1.40+: 
+- sudo apt-get install libboost-all-dev
+
+Or Boost 1.37: 
+- sudo apt-get install libboost1.37-dev
+- sudo apt-get install libqrencode-dev
```
IngCr3at1on: Not trying to nitpick but if you are doing this let's move the libqrencode-dev option up under libdb4.8 instead of under boost.

doc/build-unix.md
```diff
@@ -61,13 +61,17 @@ Versions used in this release:
...
-sudo apt-get install libssl-dev
-sudo apt-get install libdb4.8-dev
-sudo apt-get install libdb4.8++-dev
- Boost 1.40+: sudo apt-get install libboost-all-dev
- or Boost 1.37: sudo apt-get install libboost1.37-dev
-sudo apt-get install libqrencode-dev
+- sudo apt-get install build-essential
+- sudo apt-get install libssl-dev
+- sudo apt-get install libdb4.8-dev
+- sudo apt-get install libdb4.8++-dev
```
IngCr3at1on: These aren't even in the ubuntu packages last I checked so we should probably provide some information on where/how to obtain these packages.

mitchellcash: Can't we just update it with `libdb++-dev` instead?

doc/build-unix.md
mitchellcash: Agreed

mitchellcash:
I'd argue to use something like the below rather than dot points. To do this use ```. See the Code Example here https://guides.github.com/features/mastering-markdown/.
```
sudo apt-get install build-essential libssl-dev libdb++-dev libqrencode-dev
```
Boost 1.40+: 
````
sudo apt-get install libboost-all-dev
```

Or Boost 1.37: 
```
sudo apt-get install libboost1.37-dev
```

doc/build-unix.md
IngCr3at1on: I mean... I suppose we could, I know bitcoin recommends 4.8 for portability purposes but it's also particularly out of date.

If we're going to recommend a newer version we really should probably be compiling the gitian builds in the same version so that wallets can be opened by either binary.

IngCr3at1on:
I literally almost said the same thing lol

doc/build-unix.md
mitchellcash: You're right. Let's update this on how to obtain 4.8 packages.

jwrb:
I'd much prefer using code containers for this. We can even mark them as bash for correct command highlighting

RdeWilde:
No more Garza's paycoin

PR #340, opened 2015-11-05
Fix microprime maxes

IngCr3at1on:
DANGER: This is forking code, it should not be merged until a correct timestamp is set for the fork.

Currently this includes #338 and #339, it will need to be rebased before it can be merged.

PR #359, opened 2015-11-10
Microprimes stagger rates

IngCr3at1on:
As discussed in the prime-holders meeting this reduces the microprime stake rates over a 4 month period to better account for the much larger balances that microprimes can hold vs their ZC counterparts

This also changes the required balance on the microprime to match the microprime stake group at the same time as the fork.

Currently this includes #340 as it's required and should be rebased before merge.
Same with #340 this is forking code however the timestamp for when the fork occurs still needs to be set (currently it's a place holder to force the build)


dbroyhill:
Does this signal the death of Orion class staking, with MicroPrimes filling the void at an in-between rate? Shouldn't the normal wallet rate likewise be set to 20% of Prime?

IngCr3at1on:
@dbroyhill technically speaking if a wallet can/does stake it is an orion as far as the original devs coded it.

Otherwise lowering the standard wallet stake rate has been discussed, as to whether that's going to happen I'm uncertain at this time.

Issue #364, opened 2015-11-16
QT wallet allows same scape address as source address; paycoind does not


mdpfeiffer:
QT wallet allows same scape address as staking address.

By contrast, paycoind returns error message:

error: {"code":-1,"message":"Cannot set scrape address to the same as staking address."}

Not sure which one is right, or if the discrepancy could cause problems.

IngCr3at1on:
Yeah it shouldn't allow the same address, though clearly I failed to handle that on the QT side....

Issue #210, opened 2015-09-15
Scrape address UI issue


mitchellcash:
So far, in my small testing, scraping to an external address seems to be working fine but there are some display issues that might need thinking about.

In the Transactions tab we aren't really highlighting where the Stake has gone. It can look a bit confusing.

For example, in the Address column on the Transactions page we put the address from which the stake came from but don't show any info as to where the Stake has gone. This now becomes a possible hole in the transaction record.

It is possible however when you view the transaction details of a Stake to see the address where you sent the stake, however, this now becomes confusing to the user because the Address Page is showing one address where the Transaction Details are showing another address.

![](https://dl.dropboxusercontent.com/u/37902134/Paycoin%20Core/%23193_0.png)

IngCr3at1on:
I believe #247 resolves this.

IngCr3at1on:
This should be resolved, can anyone confirm?

IngCr3at1on:
Alright @mitchellcash, #247 is actually merged this time so have another look I think you'll find this is resolved.

IngCr3at1on:
@mitchellcash can you confirm if this is resolved?

mitchellcash:
Hmm I'm not sure, I would need more testing with a Standard Staking address but a quick look at my MicroPrime address on my mainnet wallet on Mac OS X I see this from 4 days ago:

![](https://dl.dropboxusercontent.com/u/37902134/GitHub/PaycoinFoundation%3Apaycoin/%23210_0.png)

But now look at https://ledger.paycoin.com/transaction/a0b76a50e7d52f48156b4cae31d28c8b1f7d924f1306e0cda93718c01410d169

You will see this was an external stake to PWeDVb9VV1Wyn2dSy1sNgGDBJPkzrGkFkQ not a mint by stake to the same address.

IngCr3at1on:
@mitchellcash an "external stake" occurs when the staking address is in a different wallet than the one that it staked to... Looks like I would have expected otherwise.

mitchellcash:
Oh right. Good point, I wrote this on my laptop in bed maybe not the best time for testing... 

In regards to same wallet scrapes outside of the scraping address I would think it should at least show the address it staked to and not the staking address. 

I would say the general rule is to always display the address the stake went to. If that is the same address that made the stake so be it, it will display that. 

I will test this weekend the actual external scraping display.

IngCr3at1on:
[...]

PR #365, opened 2015-11-16
Disallow setting a scrape address as the staking address in the QT

IngCr3at1on:
This address #364

### DIFF
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -233,6 +233,11 @@ bool AddressTableModel::setData(const QModelIndex & index, const QVariant & valu
                     return false;
                 }
 
+                /* Confirm we are not setting the scrape address to the same
+                 * address as the staking address. */
+                if (rec->address == value.toString())
+                    return false;
+
                 if (!scrapesDB->WriteScrapeAddress(rec->address.toStdString(), value.toString().toStdString()))
                     return false;
             }


jwrb:
Tested and working! It just reverts the address back to what it was.

Do you think we need some kind of error for this to inform the user that they cannot use the main address as the scrape address?

IngCr3at1on:
@jwrb that's a good question. Personally I can't imagine why anyone would ever try to scrape to their own address. I wonder what others think.

IngCr3at1on:
@mdpfeiffer

dbroyhill:
I set this once just to see what happened. There was no difference compared to not setting a scrape.

If there was a way to keep stake outputs from splitting, it might be desirable to separate the stake reward but leave it at the same address, for accounting reasons.

It's really not something I would spend any time on... as long as it doesn't break anything, who cares right now?

IngCr3at1on:
@dbroyhill see my most recent comment on #210

dbroyhill:
My conclusion is: either it should scrape as desired (same address or not), or there should be an error message.

IngCr3at1on:
Spent an hour messing with this and nothing I'm doing will actually generate a message of any kind when editing the scrape address in the GUI. It's worth mentioning it doesn't generate a message when an incorrect address is entered either.

Personally I'm for leaving this as the current behavior but it everyone would prefer we simply not block the staking address as being set as the scrape address than I'm willing to go with the consensus.

dbroyhill:
So at this time, a message of any kind for scrape errors could potentially require a lot of time with no guarantee. Current behavior is what exactly, ignore scraping to itself?

IngCr3at1on:
@dbroyhill current behavior allows setting to self in the qt but not the daemon. It causes the  reward to go to a separate output than the rest of the stake; which since it's all on the same address is a bit pointless for the fact that it will increase byte size and transaction complexity (albeit negligible at best).

Now that you mention it code could probably be put in place to ignore it during the actual transaction creation and the argument of whether to allow it wouldn't really apply lol.

dbroyhill:
My preference would be for the reward to always be a separate output, even if the scrape address is set to the staking address. If that's the current behavior, then let's leave it. If the consensus is for preventing such function, then I would be okay with the check during transaction creation to ignore it.

dbroyhill:
I successfully tested this on .3.3.0, OS X. I received the mint as a separate output to the staking address.

I like it. It might be helpful in special scenarios for accounting reasons.

In general, I prefer to err on the side of keeping unintended "features," though some would call them bugs, as long as they don't create problems. This feature adds negligible bloat to the chain - no more than a regular scrape - and doesn't cause any harm.

But again if consensus is opposed to such function, I'm okay with the transaction creation check.

IngCr3at1on:
I have no intention of directly vetoing anyone in this respect so I'd really like more feedback before we decide the best way to go with this. @CeForce you see the support tickets, what are your thoughts?

CeForce:
I haven't seen anything that makes me feel strongly one way or the other. If nothing is broken I'd say leave it alone, and as is. We haven't had too many tickets related to scrapes, if any. 

If it's disallowed without an error or some sort of feedback to the user I think it will create confusion

IngCr3at1on:
In that case the only real question is do we leave the daemon as it is or go ahead and remove the block on the daemon (it does return an error message but disallowing it on the daemon and no the qt is inconsistent).

broyhill:
As no one is strongly opposed to self scrapes, let's remove the check on the daemon.

mitchellcash:
In some ways I think we should block it with an error message like the daemon. I'm not to bothered though.

But like @IngCr3at1on says whatever we do we need to make it consistent across both qt and daemon.

IngCr3at1on:
@mitchellcash if you want to take a shot at making an error message present go for it.

PR #367, opened 2015-11-16
Check if FD_SETSIZE is already defined before trying to define it

IngCr3at1on:
If FD_SETSIZE is already set, undefine it before trying to define it

(mutes a windows warning in gitian log)

### DIFF
--- a/src/compat.h
+++ b/src/compat.h
@@ -11,6 +11,10 @@
 #ifndef NOMINMAX
 #define NOMINMAX
 #endif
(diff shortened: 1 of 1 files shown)


IngCr3at1on:
@mitchellcash want to test this on mingw (can't imagine it would cause anything but just for good measure)?

mitchellcash:
@IngCr3at1on the build was fine so I'm happy for this get in.

PR #368, opened 2015-11-16
Mute this obnoxious unused warning that occurs on linux

IngCr3at1on:


jwrb:
LGTM! Might be trivial but it's still a fix!

IngCr3at1on:
@jwrb technically it's a warning, an error would have broken the build :P

jwrb:
That's the one!

PR #371, opened 2015-11-19
Unless debugging, show a more useful format for the askfors

mitchellcash:
Will show a human readable time that the getdata will be triggered. Easier to follow.


IngCr3at1on:
Aww... you don't like unix timestamps? :P

mitchellcash:
Don't get me wrong I like them but I also like something that's more human readable haha

PR #373, opened 2015-11-19
Use getnameinfo() to get canonical IPv6 addresses

mitchellcash:


IngCr3at1on:
This works in my tests and considering it's just related to the string output I'd say it's safe.
(20 minutes)

PR #374, opened 2015-11-19
Signbugs

mitchellcash:
This pull request has fixes to prevent bit shifting signed integers to the left, onto and past the sign bit.

While this currently does not cause problems, it is undefined behavior in C++11 and could cause problems in the future (i.e. so-called time bombs). Paycoin in particular should be extra careful about these problems.


IngCr3at1on:
Tested gitian build for good measure, looks good.

PR #375, opened 2015-11-21
mapAlreadyAskedFor gets additions when AlreadyHave()

mitchellcash:
Without this change, mappings will be created even after AlreadyHave equals true, and will never be erased.

A small fix, probably would never have caused a problem, other than filling unbounded and potentially using up more memory than needed. Might be quite significant is paycoind is left running for a long period of time.


IngCr3at1on:
@mitchellcash I've been trying to track down some of the memory holes, I bet this is one lol

mitchellcash:
Let's hope so! Would love to see the memory footprint be reduced as much as possible.

PR #380, opened 2015-11-22
add client startup time as an entry to debug.log

mitchellcash:
- make ClientModel::formatClientStartupTime() return a QString

Note: Logged time in debug.log differs by a few seconds from the one displayed in the Debug window because of different initialisation times.


IngCr3at1on:
I could have sworn we did this already.

mitchellcash:
@IngCr3at1on you're thinking of #355 which was instead adding the client startup time to the GUI debug window.

IngCr3at1on:
Ahh, duh! lol

PR #386, opened 2015-11-24
Fix init issues with the qt app when running on the testnet since 284

IngCr3at1on:


jwrb:
Looks good! Works as expected! LGTM :+1:

jwrb:
This is working! LGTM! :+1:

PR #387, opened 2015-11-28
fix a memory leak in key.cpp

mitchellcash:
- add EC_KEY_free() in CKey::Reset() when pkey != NULL
- init pkey with NULL in CKey constructor


IngCr3at1on:
Hahaha, why am I not even remotely surprised by this?

mitchellcash:
You get used to it ;)

PR #389, opened 2015-11-29
Bugfix: getwork: NULL pindexPrev across CreateNewBlock, in case it fails

mitchellcash:
Without this, the following bugs are possible:

1) NULL pointer dereference

- CreateNewBlock returns NULL (out of memory); pblock is left NULL and JSON error thrown
- Next call sees pindexPrev/nTransactionsUpdatedLast/nStart up to date, so skips over making a new one
- This next call then tries to update the time on the "current" pblock (NULL)

2) Deleted object dereference

- CreateNewBlock throws an exception (not sure if this is possible right now); pblock is left with a pointer to a deleted CBlock
- Next call sees pindexPrev/nTransactionsUpdatedLast/nStart up to date, so skips over making a new one
- This next call then tried to update the time on the 'current' pblock (which is deleted)
- Consequences of this memory corruption are undefined!

### DIFF
--- a/src/bitcoinrpc.cpp
+++ b/src/bitcoinrpc.cpp
@@ -2043,8 +2043,13 @@ Value getwork(const Array& params, bool fHelp)
                     delete pblock;
                 vNewBlock.clear();
             }
+
+            // Clear pindexPrev so future getworks make a new block, despite any failures from here on
+            pindexPrev = NULL;
+
+            // Store the pindexBest used before CreateNewBlock, to avoid races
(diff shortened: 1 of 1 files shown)


scroogemcdev:
It's worth noting that this will never effect mainnet because PoW is disabled. On the offchance someone tries to use it on the testnet though or someone decides to fork the coin and enable PoW I'm OK with this (actually considered doing it before a few times I probably have it staged in a private repo lol).

PR #392, opened 2015-11-30
A few IPv6 fixes

mitchellcash:


IngCr3at1on:
Tested by trying multiple bind arguments while watching the debug log, appears to be working as it should. (30 minutes)

PR #393, opened 2015-11-30
Change strings to Paycoin (uppercase) and other miscellaneous fixes

mitchellcash:
- Update strings to use "Qt" (and not qt or QT)
- Revert a reference to Peercoin that was replaced with Paycoin
- Update initialisation of notificator to use qApp->applicationName() instead of a static string

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -461,7 +461,7 @@ void BitcoinGUI::createTrayIcon()
     trayIconMenu->addAction(quitAction);
 #endif
 
-    notificator = new Notificator(tr("p-qt"), trayIcon);
+    notificator = new Notificator(qApp->applicationName(), trayIcon);
 }
 
 #ifndef Q_OS_MAC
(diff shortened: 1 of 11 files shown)


mitchellcash:
This obviously doesn't include every correction that is out there but I wanted to at least pull a few more together compared to #353 plus a couple of other fixes so it has a better chance of getting merged :)

src/qt/bitcoingui.cpp
```diff
@@ -461,7 +461,7 @@ void BitcoinGUI::createTrayIcon()
     trayIconMenu->addAction(quitAction);
 #endif
 
-    notificator = new Notificator(tr("p-qt"), trayIcon);
+    notificator = new Notificator(qApp->applicationName(), trayIcon);
```
scroogemcdev: Purpose of this?

mitchellcash: It was updated to use a dynamic solution in favor of a static string such as "p-qt" (also this name is a bit silly you would expect something more like paycoin-qt which is a bit more friendly).

It will now get this name using qApp->applicationName() from bitcoin.cpp#L236.

It saves updating the name 'Paycoin-Qt' in multiple locations and there's no need to translate it.

PR #397, opened 2015-12-03
CBigNum: Convert negative int64 values in a more well-defined way

mitchellcash:
Since the minimum signed integer cannot be represented as positive so long as its type is signed, and it's not well-defined what happens if you make it unsigned before negating it, we instead increment the negative integer by 1, convert it, then increment the (now positive) unsigned integer by 1 to compensate

### DIFF
--- a/src/bignum.h
+++ b/src/bignum.h
@@ -131,15 +131,9 @@ public:
 
         if (sn < (int64)0)
         {
-            // We negate in 2 steps to avoid signed subtraction overflow,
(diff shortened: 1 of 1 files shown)


mitchellcash:
Sorry I was meant to include this in #374 but I must have forgotten.

IngCr3at1on:
This shouldn't be a problem but I'd like to get some extended testing on this @CeForce can you arrange that as well?

Issue #328, opened 2015-11-03
add text editor window for paycoin.conf via help menu


mdpfeiffer:
Add an option in help menu to invoke a text editor that opens paycoin.conf for easy editing.

mitchellcash:
I'll take a look at this one. I've been wanting to do it for a while.

IngCr3at1on:
This feature was implemented. Closing issue.

PR #399, opened 2015-12-03
Add a menu option to edit config file.

IngCr3at1on:
Should open the configuration file in the default text editor for the machine.
Note this has only been tested in Linux.
Addresses #328


jwrb:
This doesn't seem to open an editor for me. (OSX 10.10.5). I'll look into this further shortly

IngCr3at1on:
@jwrb most likely that's cause there is no system default editor for .conf files.

jwrb:
Double clicking config files opens them up just fine. Using the `open` command in terminal also works as expected (opening the files in sublime text)

```bash
 open ~/Library/Application\ Support/Paycoin/paycoin.conf 
```
Works just fine. So I'm not sure this is a system issue. More the way it's implemented?

jwrb:
Why did this also get merged? Didn't we merge #409 which is the same?

mitchellcash:
Hmm yes it shouldn't have been merged but because the two commit IDs are identical to the ones in #409 it actually had no ill effect.

IngCr3at1on:
@jwrb aye that was my bad but as @mitchellcash said they're identical so no ill effect... Either way, oops.

PR #400, opened 2015-12-04
Fix-up GetMyExternalIP

IngCr3at1on:
Remove dead host address for IP checking
Replace with myexternalip.com

Unfortunately myexternalip.com doesn't have an IP address so we can
only query by URL.

Changes the logic to query URLs prior to IP addresses anyway that way
if/when querying dyndns we'll use the default route for our connection
and have the IP address only if BOTH myexternalip.com and dyndns.org fail
to return properly via URL.

Switches myexternalip.com to be querried first (as the primary) that way
if USE_IPV6 is enabled (which it is by default) and the user has an
IPv6 connection (not disabled at the router/modem) they will get an IPv6
address as their local address (this is not necessarily the addrSeenByPeer
listed in getinfo (often this is an IPv4 address when using an IPv6 connection
while the internally set Local address is IPv6).

This method should be removed eventually and replaced with the current bitcoin
ip query implementation which is based entirely on peer-info; this cannot be
done easily at the moment without other changes being made to the code first
so this commit is being done to resolve the issue of relying on a single IP
checker that is guaranteed to always return an IPv4 address.

This only operates with UPNP disabled otherwise the IP address is configured
in the UPNP functionality.

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -349,7 +349,8 @@ bool GetMyExternalIP(CNetAddr& ipRet)
     if (fNoListen||fUseProxy)
         return false;
 
-    for (int nLookup = 0; nLookup <= 1; nLookup++)
+    // Reverse logic, try URL before IP address
+    for (int nLookup = 1; nLookup >= 0; nLookup--)
     for (int nHost = 1; nHost <= 2; nHost++)
     {
         // We should be phasing out our use of sites like these.  If we need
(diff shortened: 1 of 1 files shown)


jwrb:
I'll have a test of this.
I'm expecting it to work perfectly fine.

jwrb:
Tested and working! LGTM :+1:

Issue #394, opened 2015-11-30
Inconsistent White Space


jwrb:
Many files across the repo have inconsistent white space at the top of the file. This is usually under the copyright notices.

We need to have a discussion about whether we want to remove this white space, or include it in every file (for the sake of consistency).

This issue was carried down from Peercoin.

cc: @IngCr3at1on @ligerzero459 @mitchellcash

ligerzero459:
I vote for including the whitespace in every file. I believe it looks cleaner that way

IngCr3at1on:
As I mentioned in chat to @jwrb I have no strong opinion on this one way or the other but I do agree consistency would be a great idea.

mitchellcash:
Agree on consistency and side with @ligerzero459 to include the whitespace as it makes the code look cleaner.

Now to watch someone implement this commit and break everyone's build branches when it's merged :P

IngCr3at1on:
Hahaha, totally @mitchellcash

jwrb:
My personal opinion is that whitespace at the top looks better.
This would however have to be a slowly incremented change so we don't break all current PR's. Perhaps we could set some guidelines for formatting? All future PR's follow them and we can slowly work back through and sort formatting throughout the entire codebase?

IngCr3at1on:
IMHO even if it does break all outstanding PRs it's not really THAT big of an issue, any conflicts it creates will be extremely easy to resolve and being that we rarely touch the includes or copyright information it's really unlikely that it will break that much.

mitchellcash:
Yeah I don't actually mind if you do a massive breaking change. Will be some of the easiest conflicts to fix, if it even conflicts.

But yeah on the topic of styling I have my thoughts. I've actually got most of the work done in a local branch already using .clang-format (I've had it sitting there for a while now so I will have to dust it off again if we are having this discussion)

IngCr3at1on:
I was just looking at .clang-format yesterday.

mitchellcash:
What did you think? I can get a pull in

IngCr3at1on:
I have nothing against .clang-format but since it's a formatting tool I obviously wouldn't be able to say if I agree with the formatting you implemented without seeing it :P (strictly speaking I hate most of the formatting on this repo I just ignore it cause it's a hodgepodge random bits slapped together with no regards for formatting in the first place).

A formatting PR is a good place to discuss said styling though so yes, go ahead if you like.

mitchellcash:
Done, see #404

PR #404, opened 2015-12-06
Add .clang-format

mitchellcash:
This includes (requires) #403.

The idea to introduce .clang-format was based off a discussion in #394.

I have applied clang-format on a handful of files to show the effects (I didn't want to break everything by doing all files).

Command line:
```
clang-format -style=file -i coincontrol.h db.h noui.cpp addrman.cpp allocators.h mruset.h protocol.h version.cpp checkpoints.h db.cpp init.h key.h addrman.h
```


jwrb:
This formatting looks good. It's easy to read. This should be a trivial change as it should not break any features in the software.

src/allocators.h
```diff
@@ -124,7 +123,7 @@ static inline size_t GetSystemPageSize()
     page_size = sSysInfo.dwPageSize;
 #elif defined(PAGESIZE) // defined in limits.h
     page_size = PAGESIZE;
-#else // assume some POSIX OS
+#else                   // assume some POSIX OS
```
IngCr3at1on: Eww

src/db.cpp
```diff
@@ -88,14 +85,14 @@ bool CDBEnv::Open(const boost::filesystem::path& path)
...
     int ret = dbenv.open(path.string().c_str(),
-                     DB_CREATE     |
-                     DB_INIT_LOCK  |
-                     DB_INIT_LOG   |
-                     DB_INIT_MPOOL |
-                     DB_INIT_TXN   |
-                     DB_THREAD     |
-                     DB_RECOVER,
-                     S_IRUSR | S_IWUSR);
+        DB_CREATE |
```
IngCr3at1on: might have to fix this manually but I actually would prefer if all of these had the splitting pipe at the same character position.

src/db.cpp
```diff
@@ -166,28 +162,25 @@ CDB::CDB(const char *pszFile, const char* pszMode) :
...
                     throw runtime_error(strprintf("CDB() : failed to configure for no temp file backing for database %s", pszFile));
             }
 
-            ret = pdb->open(NULL,      // Txn pointer
-                            fMockDb ? NULL : pszFile,   // Filename
-                            "main",    // Logical db name
-                            DB_BTREE,  // Database type
-                            nFlags,    // Flags
-                            0);
+            ret = pdb->open(NULL,         // Txn pointer
```
IngCr3at1on: Instances like this in my opinion are actually better off as the previous instance as it lines up all the right side method data in the same position.

src/version.cpp
```diff
@@ -37,38 +37,38 @@ const std::string CLIENT_NAME("Satoshi");
 
 // First, include build.h if requested
 #ifdef HAVE_BUILD_INFO
-#    include "build.h"
+#include "build.h"
```
IngCr3at1on: I'd prefer if these were indented, pound sign and all (they're enclosed in an if statement, even if it is an ifdef).

IngCr3at1on:
I like it apart from my inline comments.

src/version.cpp
jwrb: I don't mind these. I'd have liked to see all of it indented though. 

```c
#ifdef HAVE_BUILD_INFO
    #include "build.h"
```

That would make me happy if it's possible.

jwrb:
If we chose to use clang on the entire repo; we'd have a lot to go through and sort. It'd be worth it for future. A lot of the code within Paycoin is currently messy and could do with looking smarter.

src/version.cpp
IngCr3at1on: @jwrb that's what I said :P

IngCr3at1on:
@jwrb I agree with the exception of my comments.

src/allocators.h
mitchellcash: I don't think it's that bad it's aligning with another comment.

mitchellcash:
Hmm I knew this was possibly going to be an issue with my pull but my clang-fu isn't up to scratch. I'll have to investigate those comments further but I'm not sure if it's possible or not and if we do find a fix the there is also the possibility that the fix might actually cause imperfections in other places.

Could be a lesser of two evils type commit. Let me know your thoughts before I dive in too deep...

src/allocators.h
IngCr3at1on: That's where we disagree, excessive white space is never good in my opinion. Personally I'm not for aligning trailing comments, there's literally no reason for it. Trailing comments are almost never in any way related to each other.

IngCr3at1on:
@mitchellcash that's why I generally don't rely entirely on format tools. They're great for fixing format sometimes but once you've got the bulk of it fixed there's almost always something you want to tweak manually afterwards lol

mitchellcash:
Agreed. I don't mind either way. Do we want this or not?

jwrb:
I'd still prefer right aligned pointers. @IngCr3at1on?

IngCr3at1on:
@jwrb semi-indifferent but that's how I would normally write them.

jwrb:
@IngCr3at1on Agreed. I always use right-side pointers when I'm working. It looks nicer in my opinion too.

PR #406, opened 2015-12-06
Fix regression introduced in 0b47971c918a8efadcc477e3bd16566df5957e7d

### LINKED ISSUE #405: Invalid PrimeNode Key
On the testnet using the latest master to build it seems that it is no longer correctly initiliasing using a what is a correct PrimeNode key upon a clean launch.

To reproduce:
1. Clean out the testnet3 folder apart from wallet.dat
2. In paycoin.conf ensure you are using `testnet=1` and a correct `primenodekey`
3. Will throw an error `initPrimeNode() : invalid primenode key`

If you are working from a past build in which it did allow you to use a PrimeNode key then you won't experience this issue using the latest master for whatever reason. It is only when you clean out the data directory that it starts causing this error.

mitchellcash:
Fixes #405.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -167,9 +167,6 @@ bool AppInit(int argc, char* argv[])
             return false;
         }
 
-        // Do this early so that daemon commands are processed properly
-        fTestNet = GetBoolArg("-testnet");
-
         // Command-line RPC
         for (int i = 1; i < argc; i++)
             if (!IsSwitchChar(argv[i][0]) && !boost::algorithm::istarts_with(argv[i], "paycoin:"))
@@ -339,6 +336,8 @@ bool AppInit2()
 #endif
 
     // ********************************************************* Step 2: parameter interactions
+
+    fTestNet = GetBoolArg("-testnet");
 
     if (mapArgs.count("-connect") && mapMultiArgs["-connect"].size() > 0) {
         SoftSetBoolArg("-dnsseed", false);


jwrb:
Looks good. @IngCr3at1on Could you take a look and verify this?

mitchellcash:
@jwrb sorry you don't need a prime key to test this. My issue was discovered because my prime key wasn't working on the testnet so with no investigation I posted my issue but then you will see I talk myself to a different conclusion about what is happening lol

To test this:
1. Build the latest master branch and try to run the wallet on the testnet - it won't work.
2. Now build this branch and run the wallet on the testnet - it will work.

That's about the extent of the testing needed. In any event I private messaged you a testnet prime key so you can muck around with it in testing etc.

jwrb:
Looks good then! LGTM! :+1:

IngCr3at1on:
This doesn't break command line RPC on the testnet? (which is why that setting was set where it was)?

jwrb:
I think I'd actually prefer to use #386 here

IngCr3at1on:
@jwrb I was just trying to confirm this was the same issue, though I'm pretty sure it is.

mitchellcash:
You are correct, I did only perform testing on the Qt wallet. The reason I chose my change was that it was closer to how the Bitcoin code base worked (at a a certain point in time). There will probably be disagreement over this but I try as much as I can to aim towards what Bitcoin is doing making it easier to use ideas they might have in the future (many more resources than us etc.).

But I also do not have anywhere near the knowledge of @IngCr3at1on so am happy to divert to his recommendation here.

IngCr3at1on:
@mitchellcash it's closer to how bitcoin was done yes (and how I originally did it and spent forever trying to figure out why Command-Line RPC failed), put simply that comment there is there for a reason ;)

mitchellcash:
Closed in favour of #386

PR #409, opened 2015-12-10
[FIXED] Add a menu option to edit config file.

jwrb:
Should open the configuration file in the default text editor for the machine.
Note this has only been tested on Mac. Should work on Linux
Addresses #328 and replaces PR [#399](https://github.com/PaycoinFoundation/paycoin/pull/399)

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -292,6 +292,8 @@ void BitcoinGUI::createActions()
     changePassphraseAction->setToolTip(tr("Change the passphrase used for wallet encryption"));
     openRPCConsoleAction = new QAction(QIcon(":/icons/debugwindow"), tr("&Debug window"), this);
     openRPCConsoleAction->setToolTip(tr("Open debugging and diagnostic console"));
+    editConfigAction = new QAction(QIcon(":/icons/options"), tr("&Edit config..."), this);
+    editConfigAction->setToolTip(tr("Edit the paycoin.conf file in the default text editor"));
 
     connect(quitAction, SIGNAL(triggered()), qApp, SLOT(quit()));
     connect(optionsAction, SIGNAL(triggered()), this, SLOT(optionsClicked()));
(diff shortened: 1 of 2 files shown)


src/qt/bitcoingui.cpp
```diff
@@ -950,3 +955,14 @@ void BitcoinGUI::toggleHidden()
...
+
+void BitcoinGUI::editConfig()
+{
+    QMessageBox mb;
+    mb.setWindowTitle(tr("Notice"));
+    mb.setText(tr("Any changes made to the configuration will require restarting the application."));
+    mb.exec();
+    QString configFile=QString::fromStdString(GetConfigFile().string());
+    //QDesktopServices::openUrl(QUrl(QString::fromStdString(GetConfigFile().string())));
+    QDesktopServices::openUrl(QUrl::fromLocalFile(configFile));
```
IngCr3at1on: Can this be changed back to avoid the need for the extra variable? We don't need the commented line either :D

jwrb:
I'll have a look @IngCr3at1on. It should work now. :)

ligerzero459:
Open in my default editor on OS X :+1:

jwrb:
Awesome... as soon as @IngCr3at1on can confirm it's working on for him we'll be good!

ligerzero459:
I've got time. I'll boot up my Linux VM and give it a quick test

jwrb:
Ah great! thanks!

ligerzero459:
Working like a charm on Linux as well. LGTM

IngCr3at1on:
I thought I commented already yes this works for me. Did anyone test it on windows yet?

jwrb:
I'll fire up a windows VM.

mitchellcash:
I've tested this on Windows and it works. The only nit I have is that the file type `.conf` is not used by any default application on Windows. So on first run you are presented with:



If from here you select an application from the list of installed programs and force it as your future default well then the feature works as expected.

@jwrb would you mind checking out #420.

IngCr3at1on:
@mitchellcash yeah the file association thing is really expected. Not sure there's anything we can do about that.

mitchellcash:
@IngCr3at1on I agree. I started to write down thoughts and comments but then sort of realised I had no real solution for it.

I think the benefits in this one were worth pulling into the codebase.

PR #410, opened 2015-12-10
Format QT using clang-format

jwrb:
Format the QT files using clang-format

This file is based upon @mitchellcash's PR [#404](https://github.com/PaycoinFoundation/paycoin/pull/404)
It changes the formatting in a minor way.

This is applied to the QT subdirectory as a whole. If any issues are found please report them.


scroogemcdev:
Broke it :P

jwrb:
Compiles just fine here. I don't see what the issue is?

mitchellcash:
@jwrb I think he means the fact that some changes conflict and you need to rebase over master. Which could be a risky manoeuvre in this instance so I'd just recommend revert back to master HEAD and go again. 

I guess this is why it's a good idea to only do smaller amounts of files at a time.

mitchellcash:
Also we should probably wait until my pull actually gets meted in case there are changes (that's if it actually gets merged)

IngCr3at1on:
Aye, was referring to the fact that it was conflicting. It's because of changes I made to the QT stuff somewhere in one of the PRs that was merged.

jwrb:
I will rebase this. Expect shortly

jwrb:
This has now been rebased over Master. Compiles and runs as expected with no issues. Should be ready to merge in @IngCr3at1on. I would appreciate some other eyes on this if possible to test compiling on Linux / Windows @mitchellcash @ligerzero459

mitchellcash:
I'd prefer to leave my pull/commits in #404 be the one that merges the clang-format plus fixed up docs etc.

Let's move discussion over there first and then second to that visit this as just a format pull.

PR #411, opened 2015-12-12
Small logical error in mruset

mitchellcash:
As written if you make a new_size(10) the mruset will be reduced to up to 9 elements (the while will exit when !(queue.size() >= 10) so (queue.size() < 10).

This "feature" of reducing the mruset length isn't used, so the impact on the program should be zero.

### DIFF
--- a/src/mruset.h
+++ b/src/mruset.h
@@ -51,7 +51,7 @@ public:
     size_type max_size(size_type s)
     {
         if (s)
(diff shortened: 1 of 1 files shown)


IngCr3at1on:
For some reason that commit message makes very little sense to me lol.

IngCr3at1on:
Added trivial label since this doesn't directly effect the program.

PR #412, opened 2015-12-12
Small fix to rpc_tests

mitchellcash:
sizeof(const char*) == 4 or 8 (on "sane" processor architectures) :-), sizeof(const char[]) is what you want.

As written the first substring was "04" and the second was "88c2" (on 32 bits architectures)

And I have changed the +2 in +1 so that the comment is right.

### DIFF
--- a/src/test/rpc_tests.cpp
+++ b/src/test/rpc_tests.cpp
@@ -34,9 +34,9 @@ BOOST_FIXTURE_TEST_CASE(rpc_addmultisig, TestNetFixture)
     rpcfn_type addmultisig = tableRPC["addmultisigaddress"]->actor;
 
     // old, 65-byte-long:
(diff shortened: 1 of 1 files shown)


src/test/rpc_tests.cpp
```diff
@@ -62,7 +62,7 @@ BOOST_FIXTURE_TEST_CASE(rpc_addmultisig, TestNetFixture)
     string short1(address1Hex, address1Hex+sizeof(address1Hex)-2); // last byte missing
```
IngCr3at1on: Interesting that we remove 2 here but only add 1 below...

Main problem with these tests, they check if something is thrown but have no valid way to check what (this is an issue with a lot of the tests actually).

IngCr3at1on: Yeah unless I'm mistaken that 1 comment is also wrong but otherwise this PR seems safe enough.

PR #414, opened 2015-12-12
Use linker-arguments only when linking

mitchellcash:
Passing linker-arguments when only compiling will cause warnings with Clang. This change fixes those.

Specifically this addresses warnings like these:
```
clang: warning: -Wl,-z,relro: 'linker' input unused when '-c' is present
clang: warning: -Wl,-z,now: 'linker' input unused when '-c' is present
clang: warning: argument unused during compilation: '-pie'
```

You can see the diff on my dev branch with clang turned on in Travis.
*With warnings:*
https://travis-ci.org/mitchellcash/paycoin/jobs/96497643
*Without warnings:*
https://travis-ci.org/mitchellcash/paycoin/jobs/96499171

### DIFF
--- a/src/makefile.unix
+++ b/src/makefile.unix
@@ -69,13 +69,14 @@ LIBS+= \
     # Make some important things such as the global offset table read only as soon as
     # the dynamic linker is finished building it. This will prevent overwriting of addresses
     # which would later be jumped to.
-    HARDENING+=-Wl,-z,relro -Wl,-z,now
+    LDHARDENING+=-Wl,-z,relro -Wl,-z,now
 
     # Build position independent code to take advantage of Address Space Layout Randomization
     # offered by some kernels.
     # see doc/build-unix.md for more information.
     ifdef PIE
-        HARDENING+=-fPIE -pie
+        HARDENING+=-fPIE
+        LDHARDENING+=-pie
     endif
 
     # -D_FORTIFY_SOURCE=2 does some checking for potentially exploitable code patterns in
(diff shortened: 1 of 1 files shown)


IngCr3at1on:
No need to even look at the diff, I already see those warnings in 90% of my builds lol

IngCr3at1on:
Not related to the PR but do you have any thoughts on the best way to make travis use the correct makefile for qt builds?

Currently it's building QT builds in GCC for both configs. On my local machine I can do `qmake-qt4 USE_QRCODE=1 QMAKE_CXX=clang++` to set the compiler to clang but I'm not sure the best way to have travis do it since we want it to compile with both.

mitchellcash:
@IngCr3at1on sorry for the late reply. I put this on my TODO list but missed getting to it sooner!

You want to pass `"QMAKE_CXX=$CXX"` and `"QMAKE_CC=$CC"` flags with qmake in Travis to make it work against both compilers.

You can see my Travis file: https://github.com/mitchellcash/paycoin/blob/add-travis-yml/.travis.yml
and Travis logs: https://travis-ci.org/mitchellcash/paycoin/builds/99692529

IngCr3at1on:
@mitchellcash hmm I'll check that out, though you only need CXX as we aren't using the CC compiler anywhere in the project (there is no C code only C++ :P )

mitchellcash:
Ahh you're right, well just pass the `"QMAKE_CXX=$CXX"` flag.

```
...
script:
  - qmake USE_QRCODE=1 "QMAKE_CXX=$CXX"
...
```

PR #417, opened 2015-12-13
Paycoin-Qt (Windows only): add ASLR and DEP linker flags

mitchellcash:
- for extra security on Windows: enable ASLR and DEP via GCC linker flags

GCC linker flag description:
--dynamicbase The image base address may be relocated using address space layout randomization (ASLR). This feature was introduced with MS Windows Vista for i386 PE targets.
--nxcompat The image is compatible with the Data Execution Prevention.
This feature was introduced with MS Windows XP SP2 for i386 PE targets.

Tor project is also using this.


IngCr3at1on:
This seems to build fine in gitian. Are there any possible consequences of enabling these features?

mitchellcash:
We've already had DEP on Windows builds in Paycoin since around mid August so that's fine and I also don't foresee any negatives with enabling ASLR.

IngCr3at1on:
Hmm, I'm so out of windows stuff anymore lol

I wonder if this will effect the windows 10 issues with corrupting wallets (doesn't effect 7 as far as I understand).

mitchellcash:
What's this about corrupting wallets?

IngCr3at1on:
I have very limited information as no-one who's experienced it has opened an issue but there were some mentions of windows 10 corrupting wallets.

mitchellcash:
@IngCr3at1on the issue you mention based off discussions to others is unrelated so this should be safe for merge.

IngCr3at1on:
@mitchellcash this requires rebase

mitchellcash:
@IngCr3at1on rebased

IngCr3at1on:
I don't foresee any issues with this but I'd like to get some extended testing on it anyway if possible. Can you arrange that for us @CeForce?

PR #421, opened 2015-12-13
Remove comma from setscrapeaddress description

mitchellcash:
This would confuse a user into thinking they need to put a comma in between the `<staking address>` and the `<address to stake to>` when in fact this is invalid.


jwrb:
LGTM! :+1:

PR #427, opened 2015-12-15
Only use a scrape address that is different from the staking address

IngCr3at1on:
This could use some extra testing

(30 minutes)


IngCr3at1on:
This continues on the conversation from #365 

The question now is if we want to remove the block on the daemon or leave it in place as it makes no direct difference either way?

Issue #307, opened 2015-10-28
clearorphans command does not clear orphans on QT wallet display


mdpfeiffer:
clearorphans command does not clear orphans on QT wallet display in "transactions" tab.

IngCr3at1on:
You're referring to the RPC command?

mdpfeiffer:
Yep, RPC command.

IngCr3at1on:
Yeah that's expected because QT already has a snazzy right click function that works on the transactions tab for clearing orphans from there.

Because it rebuilds that table on open, technically speaking it should remove them if you were to close and reopen the wallet after doing clearorphans in RPC.

Having said that, it may be possible to check if it's the QT wallet and call the QT function from the RPC function....

PR #428, opened 2015-12-15
Allow RPC clearorphans command to properly clear orphans in a QT wallet

IngCr3at1on:
Currently the behavior is to only call ClearOrphans in wallet.cpp
which clears the actual orphan transactions but does not remove them
from the transaction view.

This makes it so the transaction view is cleared in the event that
clearorphans is issued through RPC (or the console) on a QT wallet.


IngCr3at1on:
This addresses #307

(1 hour)

jwrb:
I'll give this a test tomorrow.

IngCr3at1on:
Rebased over master to do merge conflicts (25 minutes)

IngCr3at1on:
I'd like to get this tester and merged @CeForce and @mitchellcash

PR #429, opened 2015-12-16
A few more IPv6-related improvements

mitchellcash:


IngCr3at1on:
Tested first commit thoroughly with different bind arguments and reviewed code for second. Looks good, nice to have something other than a silent failure on a bad bind argument. (40 minutes)

PR #430, opened 2015-12-17
Change display logic for transaction table and scrape transactions

IngCr3at1on:
All scrape transactions display the address that the reward went to.

"External scrape" type is set on transactions where the staking address
is not found in the receiving wallet.

"Local scrape" type is set on transactions where the staking address
and reward/scrape address are both found in the same wallet.

"Scraped to external" type is set on transactions where the reward
address was not found in the same wallet that did the stake
In the case of such transactions the reward coins are never present in
the staking wallet and therefore are always displayed in brackets.


IngCr3at1on:
selection_190
selection_191

This addresses #210 

(3 hours 10 minutes)

IngCr3at1on:
Added a commit to reduce some repeat code (15 minutes)

IngCr3at1on:
@mitchellcash and @CeForce it would be good to get some testing on this so we can merge it as well.

PR #431, opened 2015-12-19
Qt: add copyrightLabel to aboutdialog.ui

mitchellcash:
- add a new label, which can be updated independently from the whole license information stuff
- the benefit is, we don't need to re-translate that whole wall of text every year the copyright info changes
- update to the same copyright string we use in the source and in the paycoin-qt.exe meta-data information
- removes an obsolete entry from the ui-file


IngCr3at1on:
looks good (10 minutes)

PR #432, opened 2015-12-19
Translation strings update

mitchellcash:
Base translation strings update, there have been some new strings added recently.

I will set myself a reminder to try and do this semi-regularly. Transifex will pick up changes to `paycoin_en.ts` automatically once a day.


IngCr3at1on:
looks good (25 minutes)

PR #433, opened 2015-12-20
select()'s first argument should be zero, if no file descriptors are selected

mitchellcash:
The most portable invocation for the zero-fds case, a widely used
portable hack for sleeping, generally involves passing NULL arguments
when the fd sets are unused. This form is used here.


IngCr3at1on:
Looks good but went ahead and deployed on a test node for a bit (15 minutes)

PR #435, opened 2015-12-21
Rework network config settings

mitchellcash:
The network base does no longer call GetArg directly. Instead, everything happens through a few globals and the SetProxy/SetNameProxy methods. These are called in init.cpp:AppInit2, and in Bitcoin-Qt's optionsmodel.cpp.

This includes (requires) #429.


IngCr3at1on:
Untested ack, someone should really test the proxy functionality (15 minutes)

scroogemcdev:
#429 has been merged.

mitchellcash:
Whoops! Accidentally pressed some wrong keys... and now GitHub won't accept my fixed branch (maybe because I closed this accidentally). Will open a new Pull.

PR #436, opened 2015-12-22
Rework network config settings

MitchellCash:
The network base does no longer call GetArg directly. Instead, everything happens through a few globals and the SetProxy/SetNameProxy methods. These are called in init.cpp:AppInit2, and in Paycoin-Qt's optionsmodel.cpp.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -206,13 +206,15 @@ bool static InitWarning(const std::string &str)
     return true;
 }
 
-bool static Bind(const CService &addr) {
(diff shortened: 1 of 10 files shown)


mitchellcash:
Has anyone had a chance to look at this one? I've got Tor support working as far as I can tell but it just relies on this plus a few other things I'll put in requests for once this is merged.

IngCr3at1on:
LGTM (30 minutes)

PR #366, opened 2015-11-16
Warnings signed unsigned

IngCr3at1on:
This is part of PR #229

PR #437, opened 2015-12-22
Fix sign comparison warnings

mitchellcash:
This does take some thunder from #366 but I think the commit messages are more detailed and it also includes some extra changes/improvements that #366 does not.


IngCr3at1on:
@mitchellcash so does this more or less replace #366?

mitchellcash:
@IngCr3at1on now it does ;)

src/addrman.cpp
```diff
@@ -103,12 +103,11 @@ CAddrInfo* CAddrMan::Create(const CAddress &addr, const CNetAddr &addrSource, in
     return &mapInfo[nId];
 }
 
-void CAddrMan::SwapRandom(int nRndPos1, int nRndPos2)
+void CAddrMan::SwapRandom(unsigned int nRndPos1, unsigned int nRndPos2)
 {
     if (nRndPos1 == nRndPos2)
         return;
 
-    assert(nRndPos1 >= 0 && nRndPos2 >= 0);
```
IngCr3at1on: So essentially this is meant to insure that the RndPos# variables are positive...

I'm wondering if there is any chance that they could ever be negative (if they were and this was left in place it would actually crash the application; I'm not sure what the behavior would be with this check missing); just because it was removed in Bitcoin doesn't necessarily mean it's safe or a good idea for us to remove it though...

Thoughts?

PR #439, opened 2015-12-27
Periodic translation update

mitchellcash:
- Add new translations
- update-translation script was not considering new translations
    - fixed this, also remove (nearly) empty translations


IngCr3at1on:
LGTM

PR #445, opened 2016-01-02
Rename Mac builds to Paycoin-Qt.app and add to .gitignore

mitchellcash:

PR #446, opened 2016-01-02
Update Mac Deployment Script for Paycoin

mitchellcash:
Requires #445 as I rename the Mac Application to Paycoin-Qt.app

I've tested and confirmed this working. It creates a `.dmg` like below:

Which when mounted looks like:


IngCr3at1on:
@ligerzero459 or @jwrb any chance either of you can test this at some point?

PR #450, opened 2016-01-03
Translation updates / string updates

mitchellcash:
- Split translation of RPC console welcome message and remove the need to take care of HTML br
- Remove some spaces in strings

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -418,7 +418,7 @@ bool AppInit2()
             return InitError(strprintf(_("Invalid amount for -paytxfee=<amount>: '%s'"), mapArgs["-paytxfee"].c_str()));
 
         if (nTransactionFee > 0.25 * COIN)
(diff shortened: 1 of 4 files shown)


mitchellcash:
Slowly getting there with all this translation work :) Unneeded spaces and HTML code don't bode well with the translations.

IngCr3at1on:
Aren't most of those spaces more for formatting vs "unneeded"?

mitchellcash:
I would argue they're unneeded plus they can be annoying for translators. I would rather keep it standard sentence structure with a single space following a full stop.

IngCr3at1on:
The full stops I agree but in the case of a colon followed by a blank space, that's padding so that whatever is being stuck after that colon is not right up against it. Personally I'd prefer it that way in most cases.

mitchellcash:
Right sorry. The blank space after the colon is definitely unneeded. The Qt code handles the spacing and columns itself so we don't need to add our own additional spaces as they have no affect in this particular case.

IngCr3at1on:
Got it, cool.

PR #451, opened 2016-01-10
Cleanup leftover docs from the Peerunity fork

### LINKED ISSUE #422: Clean up "testing"-folders
https://github.com/PaycoinFoundation/paycoin/tree/master/testing

I think there's some stuff herr that can be deleted. Also there are many occurences of Peercoin PPC

mitchellcash:
Fixes #422.


IngCr3at1on:
This has the only documentation of multisig to my knowledge.

mitchellcash:
You might be right. Should we put this on the Paycoin Wiki instead?

IngCr3at1on:
@mitchellcash might not be a bad idea yeah

mitchellcash:
Just a quick transposing job but it should be what you're after http://xpy.wiki/index.php?title=Multisig

(15 minutes)

PR #458, opened 2016-01-19
Save main-geometry and some header data before closing the GUI

### LINKED ISSUE #308: sticky preferences for column widths on minting tab
On the Mac, the default display for the minting tab does not allow the user to see the probability without scrolling all the way to the right. Presumably that's the estimate the user is looking for when they click the tab, so it would be nice to be able to see it straight away.

Also, the default field column widths do not correspond with the content or even the label. If the user tries to adjust the default, the wallet display returns to the defaults. It would be nice if the default settings were better tuned to Foucault the user's expectations. It would also be great if adjustments to the default were sticky preferences.

IngCr3at1on:
This will restore the physical geometry of the main window upon reopening

It will also retain and restore the column header sizes for the minting
tab and the transactions tab; do to the nature of the address book (in
that both receiving page and the address book page use the same header
object) we cannot restore this without breaking scrape address display
(I figure this is safe cause there's only 2 or 3 entries on both those
tabs and they fit pretty well in all states).

This resolves #308

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -21,6 +21,7 @@
 #include <QLocale>
 #include <QTranslator>
 #include <QLibraryInfo>
+#include <QSettings>
 
 #include <boost/interprocess/ipc/message_queue.hpp>
 #include <boost/algorithm/string/predicate.hpp>
@@ -309,6 +310,11 @@ int main(int argc, char *argv[])
             }
             // Shutdown the core and it's threads, but don't exit Bitcoin-Qt here
             Shutdown(NULL);
+
+            // Do this last to make sure all settings are written
+            QSettings settings;
+            if (settings.value("settings-version").toInt() != SETTINGS_VERSION)
+                settings.setValue("settings-version", SETTINGS_VERSION);
         }
         else
         {
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -57,6 +57,7 @@
 #include <QDragEnterEvent>
 #include <QUrl>
 #include <QStyle>
+#include <QSettings>
 
 #include <iostream>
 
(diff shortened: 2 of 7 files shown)


IngCr3at1on:
6.42 hours

mitchellcash:
Hmm I've been looking into this more and agree we should have the ability to save the main-geometry for the main GUI.

But I don't think we need to burden ourselves with saving the different column widths but I do think the columns (well the Minting Table Model) needs to work in a way that everything is visible to the user.

If we define the minimum width of the column and that minimum width shows everything visible then lets not bother saving the column width.

So my thoughts to make this happen is a change to the Minting Table Model in which we remove the Transaction column and implement the ability to double click on the transaction to open a more detailed transaction description (like in the Transaction Table).

We then make the Address column use `QHeaderView::Stretch`, reduce the MintProbability column to 120.

This should leave enough space on the Minting Table Model for everything to be viewable even on a reduced window. If the power user wishes for more detail than the behaviour should match the Transaction Table and they double click for more info.

IngCr3at1on:
@mitchellcash saving the tab states for all but the address book is trivial and already included in this commit. I see little reason not to save the tab states if we can (convenience factor and it's literally just 1 extra line in the qt generated conf file).

I'm all for adjusting the defaults but I see absolutely no reason to remove saving the tab states from this commit.

mitchellcash:
Agreed as this commit stands to allow saving adjustable column widths.

So my thinking was on a higher level to what the root cause to this was, if we make it so the whole Minting Table is viewable on Paycoin startup than there is no need to adjust columns.

And also some UX came into play which was matching the Minting tab to the Transaction tab in how it functioned.

IngCr3at1on:
@mitchellcash there may not be a need but at the same time if someone uses the wallet in a different size than the default it should be adjustable imho

Issue #309, opened 2015-10-28
display projected combined even while staking in overview tab


mdpfeiffer:
Overview tab shows balance, actively staking balance, and unconfirmed balance. Since it takes 120 confirmations for staking balance, users should still be able to see the combined account balance (projected balance post-confirmation of staking).

IngCr3at1on:
What you don't like having to add the two together? :P

IngCr3at1on:
Closing because #464 has been merged.

PR #460, opened 2016-01-20
add total (combined) balance display field to QT overview

IngCr3at1on:
Someone mentioned this at one point, I don't remember who.

I implemented it in the QT but haven't added it to the getinfo field in the RPC calls at this time.

I wanted some feedback regarding the wording for the label as well as just to know if people really want this feature and if they want it in the daemon as well.

### DIFF
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -171,14 +171,45 @@
             </property>
            </widget>
           </item>
+
           <item row="3" column="0">
+           <widget class="QLabel" name="label_7">
+            <property name="text">
+             <string>Total balance:</string>
+            </property>
+           </widget>
+          </item>
+          <item row="3" column="1">
+           <widget class="QLabel" name="labelCombinedTotal">
+            <property name="font">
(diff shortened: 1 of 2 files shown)


IngCr3at1on:
selection_198

IngCr3at1on:
(1 hour)

dbroyhill:
Nice! I think it was confusing for people new to Proof of Stake. I think more info is always better than less. As for the daemon I don't really care about it. How much work is it? Who uses the daemon and can't add? :)

IngCr3at1on:
Fixes #309 

Work wise it's almost no effort one way or the other @dbroyhill I was just trying to get an idea if people wanted said feature or not.

jpass022:
So should the first "Balance" be "Available Balance"?

dbroyhill:
Makes sense!

IngCr3at1on:
@jpass022 possibly, that's sort of why I was asking for feedback on those labels ;)

IngCr3at1on:
Closing in favor of #464

Issue #264, opened 2015-10-11
Disable upnp by default


mitchellcash:
The miniupnpc codebase seems to contain vulnerabilities and Bitcoin seems to be moving towards removing the dependency completely.

Until they can remove the dependency completely their compromise for now is to at least disable it by default, to prevent UPnP vulnerabilities being a structural danger to the network.

Can @IngCr3at1on please comment if we should also be moving towards this and if, yes, should we prioritise this work to mitigate the risk as soon as possible.

Possible work required:
- Update miniupnpc to 1.9.20151008
- Disable upnp by default

IngCr3at1on:
Yes it is a concern, I spent all of friday trying to update our gitian dependencies but succeeded in breaking the windows builds repeatedly.

IngCr3at1on:
Took another shot at trying to update the gitian descriptors to use a newer version of miniupnpc but I just get the following no matter what I do:

    ./build/net.o:net.cpp:(.text+0x7ce8): undefined reference to `__imp__upnpDiscover'
    ./build/net.o:net.cpp:(.text+0x7d20): undefined reference to `__imp__UPNP_GetValidIGD'
    ./build/net.o:net.cpp:(.text+0x7d54): undefined reference to `__imp__freeUPNPDevlist'
    ./build/net.o:net.cpp:(.text+0x7f00): undefined reference to `__imp__UPNP_AddPortMapping'
    ./build/net.o:net.cpp:(.text+0x800b): undefined reference to `__imp__UPNP_AddPortMapping'
    ./build/net.o:net.cpp:(.text+0x8022): undefined reference to `__imp__strupnperror'
    ./build/net.o:net.cpp:(.text+0x8068): undefined reference to `__imp__FreeUPNPUrls'
    ./build/net.o:net.cpp:(.text+0x8076): undefined reference to `__imp__strupnperror'
    ./build/net.o:net.cpp:(.text+0x80d3): undefined reference to `__imp__UPNP_GetExternalIPAddress'
    ./build/net.o:net.cpp:(.text+0x8186): undefined reference to `__imp__UPNP_DeletePortMapping'
    ./build/net.o:net.cpp:(.text+0x81a5): undefined reference to `__imp__freeUPNPDevlist'
    ./build/net.o:net.cpp:(.text+0x81b1): undefined reference to `__imp__FreeUPNPUrls'

The linux builds complete without issues, thought since you had windows builds working locally you might have some thoughts on this @mitchellcash 

[...]

PR #461, opened 2016-01-21
Set all builds to have UPNP disabled by default

IngCr3at1on:
this at least builds in gitian (1 hour) and partly addresses #264


jpass022:
Why would you want upnp disabled by default? I think this will just cause problems for normal users...

IngCr3at1on:
https://github.com/PaycoinFoundation/paycoin/issues/264

You'll find that even though they've updated the UPNP version for bitcoin they have disabled it be default on bitcoin as well to avoid possible issues like the above mentioned in the future.

dbroyhill:
Does disabling UPNP require users to perform additional steps, such as adding excepts to their firewalls and opening ports on their routers?

IngCr3at1on:
Possibly in some cases, otherwise they may not need to as that will only effect incomming connections. A wallet can function with nothing but outgoing connections (I've done that with a few daemons at different points).

If a user finds they have issues with the connections with UPNP disabled they can either open ports or enable UPNP in their conf or when starting the wallet. This doesn't remove UPNP it just changes it from being Opt-Out (which is potentially unsafe) to being Opt-In (requiring at least some minor education and responsibility on the users part).

CeForce:
We've had scenarios in support where we have had to help or direct users to open their firewall ports even when UPNP enabled. If it's still going to only impact the occasional or a small % of users it makes sense to set the default to UPNP disabled. 

I'd rather see something a little more inherently secure be the default and it require a user action to change that status.

IngCr3at1on:
This has been rebased over master do to merge conflicts

mitchellcash:
utACK to making it disabled by default

PR #464, opened 2016-01-24
Add total balance in overview page

mitchellcash:
This is my take on #460. Sorry @IngCr3at1on I didn't mean to come and take over but I had in mind some changes I wanted to make and I thought it easier to just do it myself so we could compare.

- The counting of the number of transactions uses different criteria than the transaction list. This has always been the case as it counts on a lower level. At least this difference exists - how to count transactions with multiple outputs. So I have removed this from the overview page.
- Changed some labels on the overview page to `Confirmed:` and `Total:`
- Add a horizontal line to improve the UX when displaying the total

It's a both an overview page cleanup together with add total balance.

(1 hour)

### DIFF
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -171,23 +171,6 @@
             </property>
            </widget>
           </item>
-          <item row="3" column="0">
-           <widget class="QLabel" name="label_2">
-            <property name="text">
-             <string>Number of transactions:</string>
-            </property>
-           </widget>
-          </item>
-          <item row="3" column="1">
-           <widget class="QLabel" name="labelNumTransactions">
-            <property name="toolTip">
-             <string>Total number of transactions in wallet</string>
-            </property>
-            <property name="text">
-             <string notr="true">0</string>
-            </property>
-           </widget>
-          </item>
          </layout>
         </item>
        </layout>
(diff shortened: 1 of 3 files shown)


IngCr3at1on:
@mitchellcash transactions with multiple outputs are still single transactions so imho that count was valid. I'd be interested to know what other people think regarding this and #460

mitchellcash:
@IngCr3at1on agreed this for the people to decide if we can't. Another reason to remove the count is that on the testnet (I know testnet) it doesn't properly count mined blocks.

Also does anyone use transaction count for anything?

Seeing as @dbroyhill and @jpass022 commented on the other code, care to comment here?

dbroyhill:
I guess I have not thought of this before but really what's the use of the transaction count? Especially if it is counted in a different way than the transactions list. Let's lose it.

I like the horizontal rule and naming, nice!

jpass022:
How about for transaction progression purposes we use the following:
Unconfirmed: On top because that's where the coin first appears
Confirmed: Second because unconfirmed coin moves to confirmed
Staking: Third because staking comes from confirmed coins
Total: Last because it is the total of everything above...

jpass022:
I think transaction count is only something I'd use for troubleshooting
purposes. I think it's fine to remove it...

dbroyhill:
@jpass022 fine with me... hopefully that makes it really clear.

IngCr3at1on:
Personally I'm fine either way. It sounds like people favor this over #460

CeForce:
I'm also for removing the tx count and adding the ui changes pictured. i think this looks good.

mitchellcash:
:+1:

IngCr3at1on:
This fixes #309

IngCr3at1on:
(30 minutes)

PR #466, opened 2016-01-25
Revert "Remove spent transactions from the minting table model."

IngCr3at1on:
This reverts commit 02005297a1f87d633deba98e690473097e1b50be.

This is causing a random double free in the memory which is crashing on some transactions (mainly in windows), revert it until we can figure out a better way to do this.

This reopens #17 
Fixes #297, fixes #456, fixes #457 and fixes #463

PR #468, opened 2016-01-26
Remove spent transactions from minting table

mitchellcash:
This builds on top of #466 and is another attempt at removing spent transactions from the minting table.

This is meant for vigorous testing before merge so we avoid the same mistakes of the past.

Would love a lot of eyes over this.


IngCr3at1on:
@mitchellcash needs rebase

mitchellcash:
@IngCr3at1on rebased

IngCr3at1on:
I've been playing with this on the testnet today (for quite awhile) and it doesn't seem to crash like my original attempt at this. Having said that the crashing was always inconsistent in Linux, @CeForce this will need to be tested on Windows.

PR #472, opened 2016-02-10
Better resolution Mac OS X icon

mitchellcash:
This always looked so gross in my dock compared to the other icons so I had to do something!

There could still be a small visual improvement but I need someone to provide me with a 1024x1024 logo... my photoshop skills aren't good enough to render my own version.

_You can figure out which is which in this image ;)_

(30 minutes)


mitchellcash:
Okay I got a copy of the 1024x1024 logo and updated this pull to use that as my base.

PR #474, opened 2016-02-10
Update testnet colours and other small graphical changes

mitchellcash:
- Change the universal testnet colour to FF7F00
- Redo testnet icon to a 256x256 square
- Redo toolbar icons to a 16x16 square
- Small update to the splash screens with slightly more resolution

Carries on some of the discussion from #227 but with the work done so if we agree we can just merge the changes.


IngCr3at1on:
@mitchellcash I'm fine with this considering the splash was already changed the icon should match @CeForce do you disagree?

PR #481, opened 2016-02-21
Replace testnet DNS Seed with a working one

### LINKED ISSUE #443: Cannot connect to testnet
My testnet wallet is failing to connect to the testnet. Is the DNS seed `tseed.paycoin.com` down maybe?

mitchellcash:
Fixes #443.

I replaced the Paycoin Testnet Seed but if there is intention to turn this node back on let me know and I will leave it in there.

(30 minutes)

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -1182,7 +1182,8 @@ static const char *strDNSSeed[][2] = {
     {"dnsseed.paycoin.com", "dnsseed.paycoin.com"},
     {"dnsseed.paycoinfoundation.org", "dnsseed.paycoinfoundation.org"},
     {"dnsseed.xpydev.org", "dnsseed.xpydev.org"},
-    {"tseed.paycoin.com", "tseed.paycoin.com"}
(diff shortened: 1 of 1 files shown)


mitchellcash:
I guess I should check with some other prominent testnet node owners as to the hashes of certain block heights... I think the testnet might be a bit like the wild west right now.

IngCr3at1on:
@mitchellcash please read the comments above the dnsseed options as a testnet seed should start with t...

You'll note that all the current seeds have both the host and the name as the same thing for consistency sake.

mitchellcash:
@IngCr3at1on done. I did read the comment but I believed the way I had it was fine.

I have also put back Yoshi's seed.

IngCr3at1on:
Thank you sir; also no the way you had it wouldn't have worked (it would have attempted to use that seed for the mainnet):
https://github.com/PaycoinFoundation/paycoin/blob/master/src/net.cpp#L1214

The first variable is the index (referred to as the seed_idx in the code).

PR #482, opened 2016-02-23
Better defined RPC and HTTP errors

IngCr3at1on:
Uses HTTP and RPC status codes from Bitcoin
Corrects inconsistencies between RPC error codes

There are a few instances instances where I was unsure whether the used error code was really valid so I commented it. In some cases error codes were changed because of inconsistencies or because of the use of the wrong error code in the first place.

This will need extensive review because of changed codes, unfortunately this may create issues for Paycoin RPC wallet implementations but should also bring some consistency between Paycoin and Bitcoin to better allow Bitcoin RPC software to work with Paycoin

(3 hours 30 minutes)

### DIFF
--- a/src/bitcoinrpc.cpp
+++ b/src/bitcoinrpc.cpp
@@ -49,10 +49,10 @@ int64 AmountFromValue(const Value& value)
 {
     double dAmount = value.get_real();
     if (dAmount <= 0.0 || dAmount > MAX_MONEY)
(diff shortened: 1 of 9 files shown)


IngCr3at1on:
Parts of this may not match bitcoin in regards to the thrown type (runtime_error vs an actual error number); this is because a runtime_error is really inappropriate for an RPC error.

PR #485, opened 2016-02-28
Correctly remove wxWidgets .exe during setup

mitchellcash:
Would someone be kind enough to make me a Windows installer file to test this :)

And when I say someone it's probably you @IngCr3at1on haha


mitchellcash:
Actually I don't think Paycoin has ever had a build with the Wx UI like Bitcoin did.

I might update this pull to remove that code instead. Thoughts?

IngCr3at1on:
Yeah I'm pretty sure that's legacy stuff from before even Peercoin.

PR #487, opened 2016-03-04
Move microprime addresses into a sub-directory and break into multipl…

IngCr3at1on:
…e files

This should resolve issues with compiling in Clang on low memory machines.


IngCr3at1on:
(6 hours (freaking build errors lol))

IngCr3at1on:
breaks windows gitian build (40 minutes)

src/makefile.mingw
```diff
@@ -127,15 +127,24 @@ OBJS= \
     obj/kernel.o \
     obj/primenodes.o \
     obj/primekeys.o \
-    obj/mpkeys.o \
-    obj/mpkeys_test.o \
-    obj/scrapesdb.o
+    obj/scrapesdb.o \
```
MitchellCash, line 130: Whoops, did we forget to add these in the original commit?

mitchellcash:
Yet to build but first comment is if we do this for MicroPrimes let's make it consistent and move PrimeNodes into a separate folder or put both in a joint folder for all Prime/MicroPrime addresses/keys?

mitchellcash:
Builds fine on Windows :) Haven't tested functionality yet

IngCr3at1on:
@mitchellcash makes little difference imho (moving the primenodes); the main reason for this is because if we break it up into multiple files like this in the main source directory it's confusing.

Need to figure out why the gitian builds are broken then...

src/makefile.mingw
IngCr3at1on, line 130: They were there, I removed them and changed them to the new files (gitian build uses linux-mingw not mingw btw).

IngCr3at1on:
Totally confused... the microprimes string shouldn't have been in the object paths for the makefiles (it's not in the unix one but I put it in the others for some reason); your windows build shouldn't have built because the object files should not have existed (or at the very least the required dir); did you do a `mkdir obj/microprimes` to force it?

IngCr3at1on:
OK. I fixed the gitian builds and the obj paths in the non-unix makefiles.

@mitchellcash if you could double checks windows builds
@jwrb or @ligerzero459 can one of y'all possibly test mac builds for this

Will squash the PR after it's fully reviewed.

(2.75 hours)

MitchellCash:
Will test this. I can also test Mac builds. 

As part of this pull though can you make the changes to .travis.yml to support Clang.

If you need tips on how to make it work I've got a working Travis file here: https://github.com/mitchellcash/paycoin/blob/add-travis-yml/.travis.yml

IngCr3at1on:
@mitchellcash the clang builds seem to be failing now, not sure why.

I can compile locally with clang without issues.

MitchellCash:
I would need to dig through the Travis documentation again but if you copy my travis.yml file verbatim clang builds will work fine.

See working branch here: https://github.com/mitchellcash/paycoin/tree/move-microprime-addresses-dev

And the working Travis build here: https://travis-ci.org/mitchellcash/paycoin/jobs/118950298

But yeah as to the reason it's something to do with Boost and the Ubuntu version Travis is building on by default so you need to enforce trusty which also requires sudo. Something like that. I can't recall the exact details.

